PR #11, opened 2021-02-01
Actions fix

lionkor:


synth80s:
FYI for debian I could not link the ppa you were using for ubuntu (the mhier one I found in your build test script). I had to download and compile the libboost1.70 binaries myself.

PR #12, opened 2021-02-02
New backend (merge when it's up)

lionkor:
Switch heartbeat URL from old one to new one, **merge this once the new backend is up.** 

Also added internal timekeeping, we now know the (real) uptime of the server for lua scripts later.

### DIFF
--- a/include/Settings.h
+++ b/include/Settings.h
@@ -6,6 +6,7 @@
 /// Created by Anonymous275 on 7/28/2020
 ///
 #pragma once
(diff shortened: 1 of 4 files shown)


Starystars67:
Could this be merged now since it is up and based on last nights conversation?

lionkor:
Yes, I'll merge it. Keep in mind that the github release action is currently broken, I will fix it tonight most likely, and then we will release as discussed. :)

PR #32, opened 2021-06-30
New Lua features

### LINKED ISSUE #35: Whitespace prepending message for the onChatMessage event.
I have found in the latest update of the server that there is prepended whitespace on the message parameter of the `onChatMessage` event in the Lua scripting interface.

Can we get this removed so that we do not need to trim the string before use?

lionkor:
Improving Lua by adding functionality, fixing some broken ones and general improvements of the codebase.


lionkor:
- Namespaced everything into `MP.*`
- `MP.HttpsGET`
- `MP.HttpsPOST`
- `MP.GetOSName`
- `printRaw`

lionkor:
Changed the table returned by `MP.GetPlayerIdentifiers(id)`.
Now, you get a table of the format (example values):
```lua
{
    beammp: "12345",
    ip: "127.0.0.1"
}
```
and so on. You see I also added IP.

lionkor:
9f646b3  doesn't compile because instead of inet_ntoa we should be using inet_ntop, that'll be fixed soon

lionkor:
Added `onShutdown` event, can be used by lua to achieve RAII (close sockets, etc.)

lionkor:
`major, minor, patch = MP.GetServerVersion()`

lionkor:
^rebase

lionkor:
- `onShutdown` event

lionkor:
Merged into and replaced by #50

PR #48, opened 2021-09-11
Actions: Prerelease by default

lionkor:
Since the Backend will change version numbers on release, we want to ensure that automated releases (on new-tag) don't trigger this mechanism. This patch makes automated releases pre-release by default, and can later manually be changed to "release". Not sure if manually changing a release from pre-release to actual release will trigger a webhook, we will have to test that. Worst-case we can just manually do real releases.


Starystars67 (approved):
Sounds good, Nice move!

PR #70, opened 2021-12-08
fix: broken .gitmodules file

RouHim:
**Current Situation**
I wanted to test the new http endpoint feature.
I noticed that the docker build failed because it could not clone some dependencies.

After looking at the `.gitmodules` file I realized why, it was broken and used ssh instead of http links.

**Expected Situation**
`.gitmodules` file is fixed.

I tried to fix it in my PR, but please have a look.


lionkor:
Works, thank you!

PR #96, opened 2022-03-17
Ignore backend response if server is Private

20dka:


lionkor:
Backend heartbeat is still sent, and any errors still reported to sentry, but no messages are printed (which get annoying and clutter the logs).

PR #123, opened 2022-09-14
Add MP.GetPositionRaw() with lua fixes and features

20dka:
add MP.GetPositionRaw(pid, vid) 

add error messages to some lua events

### DIFF
--- a/src/TServer.cpp
+++ b/src/TServer.cpp
@@ -15,6 +15,67 @@
 
 #include "Json.h"
 
+static std::optional<std::pair<int, int>> GetPidVid(std::string str) {
+    auto IDSep = str.find('-');
+    std::string pid = str.substr(0, IDSep);
+    std::string vid = str.substr(IDSep + 1);
+
+    if (pid.find_first_not_of("0123456789") == std::string::npos && vid.find_first_not_of("0123456789") == std::string::npos) {
+        try {
+        int PID = stoi(pid);
+        int VID = stoi(vid);
+        return {{ PID, VID }};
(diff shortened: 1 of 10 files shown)


src/TServer.cpp
```diff
@@ -15,6 +15,67 @@
 
 #include "Json.h"
 
+static std::optional<std::pair<int, int>> GetPidVid(std::string str) {
```
lionkor: this is performance-critical, so we want to avoid a copy here and pass by (const) ref instead
```suggestion
static std::optional<std::pair<int, int>> GetPidVid(const std::string& str) {
```

src/TServer.cpp
```diff
@@ -365,3 +424,21 @@ void TServer::InsertClient(const std::shared_ptr<TClient>& NewClient) {
     WriteLock Lock(mClientsMutex); // TODO why is there 30+ threads locked here
     (void)mClients.insert(NewClient);
 }
+
+void TServer::HandlePosition(TClient& c, std::string Packet) {
```
lionkor: avoid a copy of the entire packet string (expensive) by passing by (const) reference.
anything that isn't an int/uint/size_t/bool/ptr should be passed by const& unless semantically incorrect
```suggestion
void TServer::HandlePosition(TClient& c, const std::string& Packet) {
```

lionkor (changes requested):

lionkor:
Seriously good work, @20dka

lionkor (approved):

PR #128, opened 2022-09-26
Feature: Implement simple HTTP server to listen on the same TCP socket

lionkor:
This introduces a dependency (dlhttp) to run an http server on the existing TCP socket, which enables us to (in the future) do downloads, query server info, etc. via the same port.


lionkor:
Note that this is *unrelated* to the work on the other http server and openapi spec.

PR #134, opened 2022-10-05
Replace networking with boost::asio (synchronous)

lionkor:
- replaced all networking with boost::asio abstractions
- rewrote the SetStatus(), GetStatus() crap (now IsDisconnected() and Disconnect() with reason)
- fixed bug related to ghost players / ghost cars
- handles packets as binary for the most part (vector<uint8_t> instead of string) to fix various issues

### DIFF
--- a/.github/workflows/cmake-linux.yml
+++ b/.github/workflows/cmake-linux.yml
@@ -25,7 +25,7 @@ jobs:
         run: |
           echo ${#beammp_sentry_url}
           sudo apt-get update
-          sudo apt-get install -y libz-dev rapidjson-dev liblua5.3 libssl-dev libwebsocketpp-dev libcurl4-openssl-dev cmake g++-10
+          sudo apt-get install -y libz-dev rapidjson-dev liblua5.3 libssl-dev libwebsocketpp-dev libcurl4-openssl-dev cmake g++-10 libboost-all-dev
 
       - name: Create Build Environment
         run: cmake -E make_directory ${{github.workspace}}/build-linux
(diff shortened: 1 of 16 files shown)


lionkor:
Motivation:

The old networking was written by hand, a long time ago, and uses #ifdefs and similar magic to get winsock and posix sockets to work.

The old networking is, from my experience maintaining this for a while now:
- difficult to reason about, due to shared state (GetStatus, SetStatus)
- difficult to debug, as all platforms are different

Further, boost::asio is a very cheap abstraction layer, which also makes ipv6 support *trivial* in the future.
Another benefit is that boost::asio is documented(ish) and tested properly, whereas the previous networking code was neither documented, nor tested, or in any way properly designed.

This PR does **NOT** fix inherent design problems with the networking, those are still to-fix. Examples:
- inconsistent and undocumented packet codes (no, code is not documentation, not when its 5 functions away in some random callstack)
- strings used for packets, which breaks every time there is a `\0` byte (among other issues)
- download sockets as global(ish) state

lionkor:
To be clear: I do not have any problems understanding or reasoning about the networking code, before or after, but this makes it easier for new contributors (who we need), and makes bugs more difficult to introduce.

lionkor:
A test server with this version running on it has been at the top of the server list without issues for 220 hours, at this point. The log is 1.4 GB, because it's in TRACE + DEBUG mode, but you can believe me that there are no unexpected errors or warnings, or other issues reported.

Anonymous-275 (approved):
Clean work, Looking forward to a solid networking system.

lionkor:
Merging, the windows actions issue seems unrelated (vcpkg update)

PR #140, opened 2022-10-23
Release Candidate v3.3.0

### LINKED ISSUE #106: [Bug] MP.TriggerGlobalEvent sometimes doesn't pass arguments
OS (windows, linux, ...): Linux
BeamMP-Server Version: v3.1.0 (around 25391fa0c714c8d10ca37e2d1a74f7c2aa9cefdc)

**Describe the bug**
@0R3Z ran into this issue and reported it.

Take the following code:

This code goes in one plugin:
```lua
function bridgeWebhook(content)
    local embedTitle = content[1]
    local embedDescription = content[2]
    print("[bridgeWebhook] - " .. embedTitle .. " - " .. embedDescription)
end

function bridgeChatHandler(data)
    print("[bridgeChatHandler] 1/2 ")
    local success, err = pcall(bridgeWebhook, data)
    if not success then
        print("[bridgeChatHandler] Error: " .. err)
    end
    print("[bridgeChatHandler] 2/2 ")

end

function onInit()
    MP.RegisterEvent("bridgeChatEvent","bridgeChatHandler")
    print(vername)
    --local data = {"Beam Bridge", vername}
    --bridgeChatHandler(data) this works
    MP.TriggerLocalEvent("bridgeChatEvent", {"Beam Bridge", vername}) -- this also works
    -- but the global one doesn't work
end

vername = "Bridge V1.8 Loaded"
```

and this code goes in another plugin:
```lua
function ChatHandler(playerID,playerName, message)
    local isCommand = string.sub(message, 1, 1) == "/" --ignore commands in chat
    if isCommand then
        MP.TriggerGlobalEvent("ExecuteCommandEvent", {PlayerID, playerName, message})
        print("[ChatHandler] - isCommand")
        return 1
    else
        MP.TriggerGlobalEvent("bridgeChatEvent", {playerName, message})
        print("[ChatHandler] - isMessage")
        return 0
    end
end

function onInit()
    MP.RegisterEvent("onChatMessage","ChatHandler")
    print("Chat V1.3 Loaded")
end
```

Now, when a normal chat message is entered in game (such as `hello world`), the function `bridgeChatHandler` will *sometimes*  be called with `nil` as an argument, and sometimes with the correct data as an argument. This becomes evident in this example when passed to pcall, and subsequently indexed (`content[1]`), but the data is nil at the moment of being called.

This is not a minimally reproducible example, but small enough and likely reproducible.

### LINKED ISSUE #119: [Bug] Rapid Player Teleportation bug
OS (windows, linux, ...): Likely Both
BeamMP-Server Version: rc-v3.1.0 ? @lionkor 

**Describe the bug**
A players car will teleport rapidly between two positions on the map. Between his actual position and another of unknown origin.
This seems to affect any player that obtains a ID from the server that became "faulty". Multiple factors lead to the assumption that the now faulty ID mixes positional packets (Zp) of unknown origin (maybe from a previous player) with the actual.

All affected players have a highly fluctual ping, while they are affected. Ive gone over the issue with a affected player, he verified his game files (all clean) and reconnected with another ID and then was no longer affected. His ping was steady then.

Ive captured a instance of this bug via wireshark
https://ufile.io/vez1u3cz

The affected player in this cap is of ID 2.
When looking into the cap, you will see the server sending two Zp packets directly after another. One of which is is actual positions and the other a faulty. Also noticable is the ping difference in these packets. But generally, the structure of the packets and their content are very unusually different.
eg. https://pastebin.com/VFjqSQiP

Ive seen everyone connecting with a specific ID to be affected by this bug. Their ping was always spiking between their actualy and another from unknown origin.

A video of that can be found in the old Bug report channel of beammp (channel got archived/deleted) at the near bottom.

**To Reproduce**
Unknown

**Logs**
@lionkor

### LINKED ISSUE #151: [Bug] Server does not close after 10 seconds if there is no authkey present
OS: Windows & Linux
BeamMP-Server Version: 3.1.1

**Describe the bug**
On a fresh server, the server does not close after 10 seconds, despite it saying that it will.

**To Reproduce**
1. Start a fresh server
2. Observe it complain about the authkey
3. It does not close

**Expected behavior**
Either don't complain about the authkey, or properly close the server

lionkor:
Full changelog: <https://github.com/BeamMP/BeamMP-Server/blob/rc-v3.2.0/Changelog.md>


lionkor:
will become 3.3.0, not 3.2.0

PR #147, opened 2022-11-07
network and console improvements

20dka:
Included in this PR:
- add vehicle owner check to client packets
- update client pointer lock approach used in kick command
- add command 'quit' as alternative to 'exit'
- use boost for string lowercase comparison
- remove now unused PPSMonitor parameter

### DIFF
--- a/src/TConsole.cpp
+++ b/src/TConsole.cpp
@@ -18,6 +18,8 @@
 #include <boost/phoenix.hpp>
 #include <boost/spirit/include/qi.hpp>
 
+#include <boost/algorithm/string.hpp>
+
 static inline bool StringStartsWith(const std::string& What, const std::string& StartsWith) {
     return What.size() >= StartsWith.size() && What.substr(0, StartsWith.size()) == StartsWith;
 }
@@ -370,15 +372,15 @@ void TConsole::Command_Kick(const std::string&, const std::vector<std::string>& 
     bool Kicked = false;
     // TODO: this sucks, tolower is locale-dependent.
     auto NameCompare = [](std::string Name1, std::string Name2) -> bool {
-        std::for_each(Name1.begin(), Name1.end(), [](char& c) { c = char(std::tolower(char(c))); });
-        std::for_each(Name2.begin(), Name2.end(), [](char& c) { c = char(std::tolower(char(c))); });
-        return StringStartsWith(Name1, Name2) || StringStartsWith(Name2, Name1);
+        std::string Name1Lower = boost::algorithm::to_lower_copy(Name1);
+        std::string Name2Lower = boost::algorithm::to_lower_copy(Name2);
+        return StringStartsWith(Name1Lower, Name2Lower) || StringStartsWith(Name2Lower, Name1Lower);
     };
     mLuaEngine->Server().ForEachClient([&](std::weak_ptr<TClient> Client) -> bool {
         auto Locked = Client.lock();
         if (Locked) {
-            if (NameCompare(locked->GetName(), Name)) {
-                mLuaEngine->Network().ClientKick(*locked, Reason);
+            if (NameCompare(Locked->GetName(), Name)) {
+                mLuaEngine->Network().ClientKick(*Locked, Reason);
                 Kicked = true;
                 return false;
             }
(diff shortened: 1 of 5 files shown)


src/TConsole.cpp
```diff
@@ -369,15 +372,15 @@ void TConsole::Command_Kick(const std::string&, const std::vector<std::string>&
     bool Kicked = false;
     // TODO: this sucks, tolower is locale-dependent.
     auto NameCompare = [](std::string Name1, std::string Name2) -> bool {
```
lionkor: More a style thing than anything else, but I'd prefer never passing them as value unless we need to ;)
```suggestion
    auto NameCompare = [](const std::string& Name1, const std::string& Name2) -> bool {
```

src/TConsole.cpp
```diff
@@ -650,6 +653,69 @@ void TConsole::Command_Status(const std::string&, const std::vector<std::string>
...
+        if (name.find(stub) == 0) {
+            suggestions.push_back("lua " + name);
+        }
+    }
+}
+
+void TConsole::Autocomplete_Kick(const std::string& stub, std::vector<std::string>& suggestions) {
+    std::string stub_lower = boost::algorithm::to_lower_copy(stub);
+
+    auto LowercaseCompare = [](std::string Name1, std::string Name2) -> bool {
```
lionkor: This can likely become a function, since we use it in multiple places - for example, a `static inline` function in this file.
Either way:
```suggestion
    auto LowercaseCompare = [](const std::string& Name1, const std::string& Name2) -> bool {
```

src/TConsole.cpp
```diff
@@ -650,6 +653,69 @@ void TConsole::Command_Status(const std::string&, const std::vector<std::string>
...
+
+void TConsole::Autocomplete_Settings(const std::string& stub, std::vector<std::string>& suggestions) {
+    const std::string subcommands[] = { "help", "list", "set", "get" };
+
+    auto [command, args] = ParseCommand(stub);
+
+    std::string arg;
+    if (!args.empty()) arg = boost::algorithm::to_lower_copy(args.at(0));
+
+    auto LowercaseCompare = [](std::string Name1, std::string Name2) -> bool {
```
lionkor: See comment above - either this can be a free function here somewhere, or it needs `const&`

src/TServer.cpp
```diff
@@ -152,8 +152,11 @@ void TServer::GlobalParser(const std::weak_ptr<TClient>& Client, std::vector<uin
 
     // V to Y
     if (Code <= 89 && Code >= 86) {
-        PPSMonitor.IncrementInternalPPS();
-        Network.SendToAll(LockedClient.get(), Packet, false, false);
+        if (HandleVehicleUpdate(*LockedClient, StringPacket)){
+            Network.SendToAll(LockedClient.get(), Packet, false, false);
+        } else {
+            beammp_debugf("Invalid vehicle update packet received from '{}' ({}), ignoring it", LockedClient->GetName(), LockedClient->GetID());
+        }
```
lionkor, line 234: Nice!

src/TServer.cpp
```diff
@@ -433,24 +438,24 @@ void TServer::InsertClient(const std::shared_ptr<TClient>& NewClient) {
     (void)mClients.insert(NewClient);
 }
 
-void TServer::HandlePosition(TClient& c, const std::string& Packet) {
+bool TServer::HandlePosition(TClient& c, const std::string& Packet) {
```
lionkor: In the future it would be great to do this on the vector<>, not a string packet, since substr() and friends are super unsafe.

src/TServer.cpp
```diff
@@ -459,9 +464,45 @@ void TServer::HandlePosition(TClient& c, const std::string& Packet) {
...
+    }
+    // (Vi/We/Yl):serverVehicleID:data
+    // (Vi/We/Yl):serverVehicleID:data
+    std::string withoutCode = Packet.substr(3);
+    auto NameDataSep = withoutCode.find(':', 2);
+    if (NameDataSep == std::string::npos || NameDataSep < 2) {
+        // invalid packet
+        return false;
+    }
+    // FIXME: ensure that -2 does what it should... it seems weird.
```
lionkor: I agree this seems weird, and I'd ask to add explicit length checks before every single substr() call - if they're out of bounds, the server dies violently, so we need to be very careful (double, triple check we're not out of bounds).

lionkor (changes requested):
Amazing work, thank you very much for the contribution! I'm very excited about help like this!

I haven't run it yet, will do that later tonight or tomorrow, depending when I have time. For now, here are some nitpicks and one thing at the end I'd like us to check.

src/TServer.cpp
20dka: that chunk is a copy of the function above, but i agree it could do with more safeties

PR #161, opened 2022-12-18
fix: Bad package name for Ubuntu

rgnter:
The package name for liblua required to build is not correct. The correct name is [`liblua-5.3-dev`](https://packages.ubuntu.com/search?suite=all&searchon=names&keywords=liblua5.3)


lionkor (approved):
Just tested that, and it's correct - thank you :)

PR #165, opened 2022-12-21
[Feature Request] Password Protection

Anonymous-275:
- add hash function
- add password config
- add debug messages for password stages
- add pass Boolean for heartbeat
- adjust network codes

### DIFF
--- a/include/TNetwork.h
+++ b/include/TNetwork.h
@@ -52,4 +52,5 @@ private:
     static const uint8_t* SendSplit(TClient& c, ip::tcp::socket& Socket, const uint8_t* DataPtr, size_t Size);
 };
 
+std::string HashPassword(const std::string& str);
 std::vector<uint8_t> StringToVector(const std::string& Str);
--- a/include/TNetwork.h
+++ b/include/TNetwork.h
@@ -38,7 +38,6 @@ private:
     std::thread mUDPThread;
     std::thread mTCPThread;
 
-    static std::string Hash(const std::string& str);
     std::vector<uint8_t> UDPRcvFromClient(ip::udp::endpoint& ClientEndpoint);
     void HandleDownload(TConnection&& TCPSock);
     void OnConnect(const std::weak_ptr<TClient>& c);
--- a/include/TNetwork.h
+++ b/include/TNetwork.h
@@ -38,6 +38,7 @@ private:
     std::thread mUDPThread;
     std::thread mTCPThread;
 
+    static std::string Hash(const std::string& str);
     std::vector<uint8_t> UDPRcvFromClient(ip::udp::endpoint& ClientEndpoint);
     void HandleDownload(TConnection&& TCPSock);
     void OnConnect(const std::weak_ptr<TClient>& c);
--- a/src/TConfig.cpp
+++ b/src/TConfig.cpp
@@ -18,6 +18,7 @@ static constexpr std::string_view StrDescription = "Description";
 static constexpr std::string_view StrResourceFolder = "ResourceFolder";
 static constexpr std::string_view StrAuthKey = "AuthKey";
 static constexpr std::string_view StrLogChat = "LogChat";
+static constexpr std::string_view StrPassword = "Password";
 
 // Misc
 static constexpr std::string_view StrSendErrors = "SendErrors";
@@ -106,6 +107,7 @@ void TConfig::FlushToFile() {
     data["General"][StrMap.data()] = Application::Settings.MapName;
     data["General"][StrDescription.data()] = Application::Settings.ServerDesc;
     data["General"][StrResourceFolder.data()] = Application::Settings.Resource;
+    data["General"][StrPassword.data()] = Application::Settings.Password;
     // Misc
     data["Misc"][StrHideUpdateMessages.data()] = Application::Settings.HideUpdateMessages;
     SetComment(data["Misc"][StrHideUpdateMessages.data()].comments(), " Hides the periodic update message which notifies you of a new server version. You should really keep this on and always update as soon as possible. For more information visit https://wiki.beammp.com/en/home/server-maintenance#updating-the-server. An update message will always appear at startup regardless.");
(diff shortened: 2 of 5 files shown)


include/TNetwork.h
```diff
@@ -38,6 +38,7 @@ class TNetwork {
     std::thread mUDPThread;
     std::thread mTCPThread;
 
+    static std::string Hash(const std::string& str);
```
lionkor: This should be named more clearly, such as `HashPassword`, and probably moved to the .cpp file, if it's not called from anywhere else. We generally prefer static globals in the .cpp file to static methods in headers

src/TNetwork.cpp
```diff
@@ -235,7 +238,8 @@ std::shared_ptr<TClient> TNetwork::Authentication(TConnection&& RawConnection) {
         ClientKick(*Client, fmt::format("Invalid version header: '{}' ({})", std::string(reinterpret_cast<const char*>(Data.data()), Data.size()), Data.size()));
         return nullptr;
     }
-    if (!TCPSend(*Client, StringToVector("S"))) {
+
+    if (!TCPSend(*Client, StringToVector("A"))) { //changed to A for Accepted version
```
lionkor, line 251: This breaks all clients, doesn't it? Is this a breaking change? Clients will expect `S`, but get `A`, unless I'm misunderstanding this change

src/TNetwork.cpp
```diff
@@ -322,12 +326,30 @@ std::shared_ptr<TClient> TNetwork::Authentication(TConnection&& RawConnection) {
     }
 
     if (mServer.ClientCount() < size_t(Application::Settings.MaxPlayers)) {
+
+        if(!Application::Settings.Password.empty()) { // ask password
+            if(!TCPSend(*Client, StringToVector("S"))) {
```
lionkor: So `S` now is a *request* for password, right? We say `A` for "youre good so far", and then `S` if there's a password request? I'm just making sure this works how I read it

src/TNetwork.cpp
```diff
@@ -322,12 +326,30 @@ std::shared_ptr<TClient> TNetwork::Authentication(TConnection&& RawConnection) {
     }
 
     if (mServer.ClientCount() < size_t(Application::Settings.MaxPlayers)) {
+
+        if(!Application::Settings.Password.empty()) { // ask password
```
lionkor: Password check **needs** to happen before `onPlayerAuth`, because `onPlayerAuth is supposed to be called only once everything except player count is checked - so this needs to be changed (moved to 285). We want the client to cause as little state change as possible.

lionkor (changes requested):
Please clarify / change the requested parts. The rest looks good to me (but I was not yet able to test it - I'd ask others to do that).

lionkor:
To be clear - im asking about the specific protocol, like what & when (S, A, etc) because, if this breaks older networking, we'd have to call this a major version release

lionkor:
And, for the future, start each commit with just the action verb (like `add`, no `-` infront)

include/TNetwork.h
Anonymous-275: but you already have static methods right below it, and the method just hashes a string it can be anything not just the password. But sure I can rename it to "SHA256" if you want.

src/TNetwork.cpp
Anonymous-275: nope, this is just a byte to make the client continue it doesn't check it just wait for any signal

src/TNetwork.cpp
Anonymous-275: yes

src/TNetwork.cpp
Anonymous-275: I would beg to disagree, you authenticate then you know who you are dealing with, it adds extra security and you know who failed to connect by name

Anonymous-275:
> To be clear - im asking about the specific protocol, like what & when (S, A, etc) because, if this breaks older networking, we'd have to call this a major version release

it doesn't I obviously wouldn't break backwards compatibility and laugh while doing so

include/TNetwork.h
lionkor: It hashes a password, so it should be named accordingly, I would say. It doesnt matter how, its just about what intend to do with it

src/TNetwork.cpp
lionkor: So it doesnt care if it's A or S? What is the idea then with making it send A

src/TNetwork.cpp
lionkor: You are doing the password check after calling onPlayerAuth, which will break existing plugins, please move it to the desired location.

onPlayerAuth assumes that it is the last step in authentication, and that if it succeeds, the player will join if the player limit allows it. 

The player should not be made known to lua before passing the password check, as this already introduces the player to the system - we want to block the player as early as possible.

lionkor: Make sure you read again which line in the code im referencing - im saying to move it after authentication, before the call to onPlayerAuth

lionkor:
I just need clarification on how exactly the new protocol differs from the old one, in reasonable detail, and I seem to not be understanding it right. Could you please make clear exactly what the protocol is before and after here? Only the Authentication part

include/TNetwork.h
lionkor: To be clear, my suggestion is

- name it HashPassword
- move it to the .cpp file

src/TNetwork.cpp
Anonymous-275: It needs to wait for the server to finish authentication obviously otherwise there would be no point of continuing it just waits for anything back other than a disconnect

src/TNetwork.cpp
Anonymous-275: So since you were not clear enough you mean check before lua auth call but after the actual auth check against the backend?

lionkor: Not sure what the issue is, move it to around line 285 as i said in my original review

lionkor:
I believe everything should be clear now, please re-request a review once the changes are implemented

src/TNetwork.cpp
Anonymous-275: ok then I will blindly put it where you want then

lionkor:
You can save yourself the snarky comments. I'll review it, thank you

Anonymous-275:
> You can save yourself the snarky comments. I'll review it, thank you

what? ok

src/TConfig.cpp
```diff
@@ -106,6 +107,7 @@ void TConfig::FlushToFile() {
     data["General"][StrMap.data()] = Application::Settings.MapName;
     data["General"][StrDescription.data()] = Application::Settings.ServerDesc;
     data["General"][StrResourceFolder.data()] = Application::Settings.Resource;
+    data["General"][StrPassword.data()] = Application::Settings.Password;
```
lionkor, line 110: let's do this also, I feel this would be nice
```suggestion
    data["General"][StrPassword.data()] = Application::Settings.Password;
    SetComment(data["General"][StrPassword.data()].comments(), " Sets a password on this server, which restricts people from joining. To join, a player must enter this exact password. Leave empty ("") to disable the password.");
```

lionkor (approved):
Looks good - I added one suggestion, which you can put in if you feel it makes sense.

PR #188, opened 2023-07-30
Add dependencies for Fedora

externalxdpro:
These are the packages that were required for me to successfully build from source on Fedora 38.


lionkor:
Hi, thanks for the contribution. We've changed our build system, you should now only need liblua

PR #192, opened 2023-09-28
Added Tags Feature to the BeamMP Server. 

Starystars67:
This PR adds a tags functionality to the BeamMP server config and heartbeat.
The intended purpose of this is to enable better filtering on the server list based on keywords such as gamemode or roleplay.

### DIFF
--- a/include/Common.h
+++ b/include/Common.h
@@ -137,6 +137,8 @@ private:
     static inline Version mVersion { 3, 2, 0 };
 };
 
+void SplitString(std::string const& str, const char delim, std::vector<std::string>& out);
+
 std::string ThreadName(bool DebugModeOverride = false);
 void RegisterThread(const std::string& str);
 #define RegisterThreadAuto() RegisterThread(__func__)
--- a/include/Common.h
+++ b/include/Common.h
@@ -43,7 +43,7 @@ public:
     struct TSettings {
         std::string ServerName { "BeamMP Server" };
         std::string ServerDesc { "BeamMP Default Description" };
-        std::string ServerTags { "BeamMP,Server" };
+        std::string ServerTags { "Freeroam" };
         std::string Resource { "Resources" };
         std::string MapName { "/levels/gridmap_v2/info.json" };
         std::string Key {};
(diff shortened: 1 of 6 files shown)


lionkor:
TOML has arrays, so this should probably be an array:
```
Tags = ["Tag1", "Tag2"]
```

Starystars67:
Hi Lion, 

I have been thinking this one over over the last week and bit and I am thinking that though for us more technically capable people an array is the correct approach. 

BeamMP however is also used my people who are not as technically capable and I worry that a potential missing quote mark or comma in an array could cause the config to not be loaded and then raise new posts in support.

Happy to hear other sides on this though.

lionkor:
Fair enough, it's definitely a more complicated syntax as a TOML array. What bothers me about this approach is that it's an undocumented syntax, the parser for which will be in the backend. In this case, the comment should probably explain the syntax, something like "Separate tags with commas, like so: Tag1,Tag2,Tag3" or something.
And then any tool parsing or writing to the config will have to also write a parser/serializer for it, so either way would be good to document it in the file itself.

Other than that, maybe the team can decide on some good default tags, because I can't think of any.

As for the windows compile error, that maybe a dependency issue, Anon might be able to help there.

Starystars67:
> Fair enough, it's definitely a more complicated syntax as a TOML array. What bothers me about this approach is that it's an undocumented syntax, the parser for which will be in the backend. In this case, the comment should probably explain the syntax, something like "Separate tags with commas, like so: Tag1,Tag2,Tag3" or something. And then any tool parsing or writing to the config will have to also write a parser/serializer for it, so either way would be good to document it in the file itself.

Yep, that is what I did on line 106 in src/TConfig.cpp.
 
> Other than that, maybe the team can decide on some good default tags, because I can't think of any.

Indeed.

> As for the windows compile error, that maybe a dependency issue, Anon might be able to help there.

Will do. @Anonymous-275 :P

lionkor:
Needs better or no default tags

Starystars67:
> Needs better or no default tags

Done. Changed the default to Freeroam to match the majority of servers presently on the list.

Starystars67:
Backend support for this is already implemented.

lionkor:
Blocked by #227

lionkor:
Made the tags pretty-print in the debug view

image

image

lionkor (approved):

PR #193, opened 2023-11-26
Use native setsockopt to set the socket receive timeout

### LINKED ISSUE #171: [Bug] Joining Server fails
Since #165 was merged joining a server causes the following error:

```
[16/02/23 19:22:26] [ERROR] fatal: set_option: Invalid argument [system:22 at /usr/include/boos      t/asio/detail/reactive_socket_service.hpp:174 in function 'set_option']                              
[16/02/23 19:22:29] [ERROR] fatal: set_option: Invalid argument [system:22 at /usr/include/boos      t/asio/detail/reactive_socket_service.hpp:174 in function 'set_option']                              
[16/02/23 19:23:28] [ERROR] fatal: set_option: Invalid argument [system:22 at /usr/include/boos      t/asio/detail/reactive_socket_service.hpp:174 in function 'set_option'] 
```

Dreae:
This updates TNetwork.cpp to use the platform native `setsockopt` instead of boost `set_option` to set the receive timeout for sockets. fixes #171 

Tested locally on windows and linux

### DIFF
--- a/src/TNetwork.cpp
+++ b/src/TNetwork.cpp
@@ -11,8 +11,6 @@
 #include <boost/asio/ip/address_v4.hpp>
 #include <cstring>
 
-typedef boost::asio::detail::socket_option::integer<SOL_SOCKET, SO_RCVTIMEO> rcv_timeout_option;
-
 std::vector<uint8_t> StringToVector(const std::string& Str) {
     return std::vector<uint8_t>(Str.data(), Str.data() + Str.size());
 }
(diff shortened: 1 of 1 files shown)


src/TNetwork.cpp
```diff
@@ -153,7 +151,8 @@ void TNetwork::TCPServerMain() {
             if (ec) {
                 beammp_errorf("failed to accept: {}", ec.message());
             }
-            ClientSocket.set_option(rcv_timeout_option{ 120000 }); //timeout of 120seconds
+            const int timeout = 120000; //timeout of 120seconds
+            ::setsockopt(ClientSocket.native_handle(), SOL_SOCKET, SO_RCVTIMEO, reinterpret_cast<const char*>(&timeout), sizeof timeout);
```
lionkor, line 160: `setsockopt` on windows expects a `const char*`, on linux a `const void*`. So far, so good.
Hi, thanks for your contribution!

The WinSock API says the argument should be `milliseconds` (so 120000 is right), whereas the linux api says it expects a `timeval` struct (https://www.man7.org/linux/man-pages/man3/timeval.3type.html) which is a struct with seconds and µs.
Not sure how you tested it, but I don't see that working, since 120000 will be most likely be "understood" as the seconds in the struct, and the µs will be UB to access since there's nothing there.

I'm aware that the current timeout in the code before also doesn't work.

lionkor (changes requested):

src/TNetwork.cpp
Dreae: 'Tested' in the sense that I just compiled it for both platforms and made sure the server started and was joinable. 

I've updated this PR to pass milliseconds to `setsockopt` on windows and a `timeval` struct on other platforms (OS X also expects a timeval when setting `SO_RCVTIMEO`)

lionkor:
Solved differently by rewriting the protocol lol

PR #197, opened 2023-12-05
New buildsystem

lionkor:


snepsnepsnep (approved):

RouHim:
@lionkor Do I need to update the server build the container image? Or is it still the same? Anything I need to consider?

lionkor:
Probably, yeah - not sure what it did before, but the entire build system is now different (simpler).

Check out the `scripts/` folder, it has some helper scripts. If you just run scripts 1, 2, and 3 in a container, you end up with a working build. You run the scripts from the root of the repo.

lionkor:
You can also entirely ignore the flags in the script 2 (the CXX_ whatever flags) if you wanna instead just copy out commands.

RouHim:
nvm

PR #198, opened 2023-12-05
Fix lua assertions crashing the server

### LINKED ISSUE #222: Don't let Lua assertions kill the server
Currently, Lua assertions terminate the server. These are usually fatal, so this is technically right, but usually you want your server to stay online regardless. This pretty much never happens, but when it does, at least it should be logged and not terminate.

lionkor:
The server would crash on a Lua plugin ASSERT() - this now instead prints an error. May still lead to weird behavior, but less weird than a crash.

### DIFF
--- a/include/TLuaEngine.h
+++ b/include/TLuaEngine.h
@@ -18,6 +18,9 @@
 #include <vector>
 
 #define SOL_ALL_SAFETIES_ON 1
+#define SOL_USER_C_ASSERT SOL_ON
+#define SOL_C_ASSERT(...) \
+    beammp_lua_errorf("SOL2 assertion failure: Assertion `{}` failed in {}:{}. This *should* be a fatal error, but BeamMP Server overrides it to not be fatal. This may cause the Lua Engine to crash, or cause other issues.", #__VA_ARGS__, __FILE__, __LINE__)
 #include <sol/sol.hpp>
 
 using TLuaStateId = std::string;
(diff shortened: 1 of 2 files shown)


src/main.cpp
```diff
@@ -89,7 +89,7 @@ int BeamMPServerMain(MainArguments Arguments) {
         Application::Console().WriteRaw("BeamMP-Server v" + Application::ServerVersionString());
         return 0;
     }
-
+
```
Starystars67: Random whitespace.

include/TLuaEngine.h
```diff
@@ -18,6 +18,9 @@
 #include <vector>
 
 #define SOL_ALL_SAFETIES_ON 1
+#define SOL_USER_C_ASSERT SOL_ON
+#define SOL_C_ASSERT(...) \
+    beammp_lua_errorf("SOL2 assertion failure: Assertion `{}` failed in {}:{}. This *should* be a fatal error, but BeamMP Server overrides it to not be fatal. This may cause the Lua Engine to crash, or cause other issues.", #__VA_ARGS__, __FILE__, __LINE__)
```
Starystars67, line 23: To confirm, will this print out the file and line where the assert was called to assist the user in debugging & fixing / reporting?

lionkor: It prints out the file and line in the C++, not the Lua -- sadly an assert() means that there was a pretty much fatal error inside the lua engine, so yeah, the user should just report it.

PR #199, opened 2023-12-05
Remove password setting from config to avoid confusion

lionkor:
...until implemented.

The client afaik currently doesn't support this setting.


Lucky4Luuk (approved):
Looks good to me!

PR #200, opened 2023-12-05
Update readme to new build system

lionkor:
Now building is like 2 steps, so the readme reflects that now.

### DIFF
--- a/README.md
+++ b/README.md
@@ -43,7 +43,7 @@ We only allow building unmodified (original) source code for public use. `master
 
 ## Supported Operating Systems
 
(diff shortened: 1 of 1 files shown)


Lucky4Luuk (changes requested):
Hey, this change seems pretty good to me, but there's a few notes I have:
- You have added build instructions, but the old build instructions still seem to be there. They are listed right underneath the runtime dependencies.
- The new build instructions don't list the requirements for building on Windows. I think this is worth writing down?
- In "Supported Operating Systems", it still links to the Prerequisites section, which is missing.

Starystars67:
The points @Lucky4Luuk makes here are very good, Especially the steps for windows as that would be me :)

lionkor:
Sorry @Lucky4Luuk if its not clear, but the windows build works the same. That was most of the point of this rewrite of the build system - ill fix the point about the old instructions.

lionkor:
So @Starystars67 the same steps on windows, you just open the cmakelists with visual studio and it should just work

Lucky4Luuk (approved):
Looks good to me now, I have no more comments!

Lucky4Luuk (approved):

PR #215, opened 2023-12-21
Fix exit after 10 seconds

### LINKED ISSUE #151: [Bug] Server does not close after 10 seconds if there is no authkey present

aeroc7:
Found this bug when I was creating a server of my own. Hopefully this closes #151 indefinitely.

### DIFF
--- a/include/Common.h
+++ b/include/Common.h
@@ -75,7 +75,7 @@ public:
     static void RegisterShutdownHandler(const TShutdownHandler& Handler);
     // Causes all threads to finish up and exit gracefull gracefully
     static void GracefullyShutdown();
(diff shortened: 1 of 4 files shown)


lionkor:
Works for me, but doesn't close quite yet because it waits for console input for me. I'll see what I can do

lionkor (approved):
lgtm

PR #218, opened 2023-12-21
Remove [Bug] title from issue template

lionkor:


.github/ISSUE_TEMPLATE/bug_report.md
```diff
@@ -1,7 +1,6 @@
 ---
 name: Bug report
 about: Create a report to help us improve
-title: "[Bug] Enter issue title here"
```
Starystars67, line 4: I wonder could just keeping [Bug] be better?
This way it can be easily seen in the list? 
Maybe I misunderstand what title means in this context?

lionkor: My thinking was that we have tags to show something is a bug, so the [Bug] is like a tag that is sometimes, but not always, there as well. I focus on tags on this repo, because they're consistent and people tend to not mess with them as much as with titles :)

lionkor: image

and this works better than visually filtering for me, and i dont consider bugs more or less important than features

Starystars67: Fair enough then :)

Will approve now.

Starystars67 (approved):

PR #220, opened 2023-12-21
Fix GitHub Actions not downloading the latest tests to run them

### LINKED ISSUE #219: Run-tests actions not downloading cross-workflow artifacts
Currently, the github action for running tests doesn't run, because somehow the download-artifact action doesn't allow downloading artifacts from a different workflow.. or something. https://github.com/actions/download-artifact/issues/3. Literally the only thing you'd use this for, but whatever.

Fix by using https://github.com/dawidd6/action-download-artifact

Edit: Might be a different issue!

lionkor:
Closes #219

### DIFF
--- a/.github/workflows/linux.yml
+++ b/.github/workflows/linux.yml
@@ -50,11 +50,11 @@ jobs:
       - name: Build Tests
         run: bash ./scripts/debian-11/3-build-tests.sh
 
-      - name: Archive server tests artifact
-        uses: actions/upload-artifact@v2
-        with:
-          name: BeamMP-Server-debian-tests
-          path: ./bin/BeamMP-Server-tests
+      - name: Install Runtime Dependencies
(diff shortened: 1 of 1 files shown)


Starystars67 (commented):
Reviewing this I notice that you removed the debian and ubuntu tests and only now have the test for ubuntu. 

Was this intended? If so how are debian builds and tests now done?

lionkor:
No idea why the diff looks so weird, it's very odd. But if you go to the actual file, it has tests on both, unless im missing something there

Starystars67:
Indeed I see that now when viewing the complete file.

To be clear, I understand that the tests are a part of the build process now correct?

Starystars67 (approved):

lionkor:
Yes, the builds will fail if the tests fail -- that's not *great*, but really, if the tests fail, something is fucked royally. The unit tests currently unit test only some small parts, but still!

PR #224, opened 2023-12-22
Persist http connections where possible

lionkor:


lionkor:
Probably doesn't make sense to do it here, like this, i'll make an issue instead.

PR #236, opened 2023-12-28
Add self update commandline flag

### LINKED ISSUE #233: Add self-upgrade commandline flag
- [x] Add --update flag
- [x] Fail update if platform binary is not provided (fail based on linux distro, OS, architecture, distro version)
- [x] Pull binaries from latest release
- [x] Get latest release tag from github's API
- [ ] Explicit success message
- [ ] Add server command `self-update`
- [ ] Investigate again if self update can restart the server automatically

lionkor:
You can now update the server via an `--update` commandline flag

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -48,6 +48,7 @@ set(PRJ_HEADERS
     include/TScopedTimer.h
     include/TServer.h
     include/VehicleData.h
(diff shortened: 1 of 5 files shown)


lionkor:
This should be more explicit about success at the end, and also generate the log file and write everything there

PR #240, opened 2023-12-29
Disable server config generation via ENV

### LINKED ISSUE #238: Disable ServerConfig Generation
I would like to use ENVs to set config values for my BeamMP server. As a result, some people who are unfamiliar with this config method may still try to change the config. As such, I would like to put a warning in this file, but the server will overwrite it.

An ENV config value to disable the generation of the config file would be great, so I can put things in there myself.

lionkor:


lionkor:
waits for #239

lionkor:
but can be reviewed ;)

lionkor:
low prio

PR #245, opened 2023-12-30
Update linux.yml to build ARM64 binaries for debian11

lionkor:


Starystars67 (approved):
All looks good to me.
Testing GH actions is a bit harder. Wish they would provide an automated test that would check them.

Issue #225, opened 2023-12-24
Ungraceful disconnect does not call onVehicleDeleted


lionkor:
When the player Alt+F4 (or other ungraceful disconnect) quits, the onVehicleDeleted is not (always?) called on all the player's vehicles.

I believe that, currently, the onVehicleDeleted is not fired on car delete *at all*, and instead only on vehicle delete packets.

PR #251, opened 2024-01-08
Major refactor of Client and Server

lionkor:
this refactor includes changes to TClient:

- all member fields are now public, but protected with Sync (an alias for boost::synchronized_value
- removed all (now) obsolete getters and setters

changes to TServer and TNetwork:

- thread-safe ID generation, previously it was possible for there to be ID duplicates. this is now solved by moving id generation and assignment into the same mutex locked context.
- replaced ForEachClientWeak with ForEachClient, getting rid of the weak_ptr shit in most places
- implemented a bunch of new functions for getting rid of more weak_ptr everywhere

### DIFF
--- a/include/Client.h
+++ b/include/Client.h
@@ -10,6 +10,8 @@
 #include "BoostAliases.h"
 #include "Common.h"
 #include "Compat.h"
+#include "RWMutex.h"
+#include "Sync.h"
 #include "VehicleData.h"
 
 class TServer;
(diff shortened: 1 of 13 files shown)


lionkor:
#225 is also being closed here, among hopefully some other bugs, needs testing ofc

lionkor:
obsoleted by new protocol i guess

PR #267, opened 2024-01-23
Add Lua execution profiler `Util.DebugExecutionTime()`

### LINKED ISSUE #264: Lua event execution profiler
When I write a plugin, and it seems to slow down how responsive the server is. I can't tell what takes how long!

I would like to get a quick overview of execution times of my event handlers. I would like those to happen on command (via a console command, for example, in the `lua` console for my plugin, for example), and via a lua api (like MP.DebugExecutionTimes() or something). Once I have those times, I would like to be able to edit the code of the worst offending/most surprisingly bad performing event handlers to include manual profiling steps, which are added to the same profile.

lionkor:
Adds `Util.DebugExecutionTime()`, which returns a table of `function_name: milliseconds`, in which each function's execution time is averaged over all time.

You can call this function like so:
```lua
-- event to print the debug times
MP.RegisterEvent("printStuff", "printStuff")
-- prints the execution time of all event handlers
function printStuff()
	print(Util.DebugExecutionTime())
end
-- run every 5 seconds (or 10, or 60, whatever makes sense for you
MP.CreateEventTimer("printStuff", 5000)
```

Pretty print function:
```lua
function printDebugExecutionTime()
    local stats = Util.DebugExecutionTime()
    local pretty = "DebugExecutionTime:\n"
    local longest = 0
    for name, t in pairs(stats) do
        if #name > longest then
            longest = #name
        end
    end
    for name, t in pairs(stats) do
        pretty = pretty .. string.format("%" .. longest + 1 .. "s: %12f +/- %12f (min: %12f, max: %12f) (called %d time(s))\n", name, t.mean, t.stdev, t.min, t.max, t.n)
    end
    print(pretty)
end
```

`Util.DebugExecutionTime()` returns a table, where each key is an event handler function  name, and each value is a table consisting of `mean` (simple average), `stddev` (standard deviation aka mean of the variance), `min` and `max`, all in milliseconds, as well as `n` as the number of samples taken.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -49,6 +49,7 @@ set(PRJ_HEADERS
     include/TServer.h
     include/VehicleData.h
     include/Env.h
+    include/Profiling.h
 )
 # add all source files (.cpp) to this, except the one with main()
 set(PRJ_SOURCES
@@ -72,6 +73,7 @@ set(PRJ_SOURCES
     src/TServer.cpp
     src/VehicleData.cpp
     src/Env.cpp
+    src/Profiling.cpp
 )
 
 find_package(Lua REQUIRED)
--- a/include/Profiling.h
+++ b/include/Profiling.h
@@ -45,6 +45,9 @@ struct UnitProfileCollection {
     /// Returns the number of elements the moving average is calculated over.
     size_t measurement_count(const std::string& unit);
 
+    /// Returns the averages for all stored units.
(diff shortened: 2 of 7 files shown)


lionkor:
Now using a running calculation by using two running sums, which means we can get the mean and stdev over any number of samples.

lionkor:
Carp has been testing this extensively

PR #274, opened 2024-02-07
Add `version` command

### LINKED ISSUE #148: [Feature Request] Add `version` console command to display the server version
It would make it easier to maintain and update the server if I could easily check the version with a command. Software like Pterodactyl may not allow you to scroll up far in the console

a `version` command in the server console

Displaying the version number in the `status` command

sla-ppy:
closes #148

### DIFF
--- a/src/TConsole.cpp
+++ b/src/TConsole.cpp
@@ -248,7 +248,7 @@ void TConsole::Command_Help(const std::string&, const std::vector<std::string>& 
         settings [command]      sets or gets settings for the server, run `settings help` for more info
         status                  how the server is doing and what it's up to
         clear                   clears the console window
-        version                 return current version to console)";
+        version                 displays the server version)";
     Application::Console().WriteRaw("BeamMP-Server Console: " + std::string(sHelpString));
 }
 
@@ -273,7 +273,7 @@ void TConsole::Command_Version(const std::string& cmd, const std::vector<std::st
         return;
     }
 
-    mCommandline->write("Current version: " + Application::ServerVersionString());
+    Application::Console().WriteRaw("Current version: " + Application::ServerVersionString());
 }
 
 void TConsole::Command_Kick(const std::string&, const std::vector<std::string>& args) {
--- a/src/TConsole.cpp
+++ b/src/TConsole.cpp
@@ -273,7 +273,7 @@ void TConsole::Command_Version(const std::string& cmd, const std::vector<std::st
         return;
     }
 
-    Application::Console().WriteRaw("Current version: " + Application::ServerVersionString());
+    Application::Console().WriteRaw("Current version: v" + Application::ServerVersionString());
 }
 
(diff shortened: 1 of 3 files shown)


src/TConsole.cpp
```diff
@@ -247,7 +247,8 @@ void TConsole::Command_Help(const std::string&, const std::vector<std::string>&
         lua [state id]          switches to lua, optionally into a specific state id's lua
         settings [command]      sets or gets settings for the server, run `settings help` for more info
         status                  how the server is doing and what it's up to
-        clear                   clears the console window)";
+        clear                   clears the console window
+        version                 return current version to console)";
```
lionkor: Would be better if the wording here more closely matched the other help strings, like "displays this help", perhaps like:
```
        version                 displays the version of this server)";
```

src/TConsole.cpp
```diff
@@ -267,6 +268,14 @@ void TConsole::Command_Clear(const std::string&, const std::vector<std::string>&
     mCommandline->write("\x1b[;H\x1b[2J");
 }
 
+void TConsole::Command_Version(const std::string& cmd, const std::vector<std::string>& args) {
+    if (!EnsureArgsCount(args, 0)) {
+        return;
+    }
+
+    mCommandline->write("Current version: " + Application::ServerVersionString());
```
lionkor: For consistency, we could use `Application::Console().WriteRaw()` or just `WriteRaw()` instead... Not sure why the other functions are inconsistent about the use of this as well.

Maybe just changing this to use `Application::Console().WriteRaw()` might be nice, to at least be consitent with most of the other commands. A separate issue can be opened to address this inconsistency and the weird way in which it's called (via Application::?? why???? who knows.)

lionkor (changes requested):

lionkor:
Thanks for your contribution! Please see if you can make the requested changes, and then request a review from me again :)

sla-ppy:
@lionkor Please review my changes so the issue can be finalized.

src/TConsole.cpp
```diff
@@ -267,6 +268,14 @@ void TConsole::Command_Clear(const std::string&, const std::vector<std::string>&
     mCommandline->write("\x1b[;H\x1b[2J");
 }
 
+void TConsole::Command_Version(const std::string& cmd, const std::vector<std::string>& args) {
+    if (!EnsureArgsCount(args, 0)) {
+        return;
+    }
+
+    Application::Console().WriteRaw("Current version: " + Application::ServerVersionString());
```
lionkor: could we get a:
```suggestion
    Application::Console().WriteRaw("Current version: v" + Application::ServerVersionString());
```
that way its consistent with the other version printouts

lionkor (changes requested):
pls change this as well :)

lionkor (approved):
lgtm

PR #277, opened 2024-02-08
Fix assigment of ID's so two unique users dont get the same ID.

### LINKED ISSUE #154: [Bug] TNetwork::OpenID may assign the same ID to two players
If two players both call TNetwork::OpenID at the same time, the function may assign them both the same ID. This can be fixed by making the calls exclusive to each other in some way.

sla-ppy:
Closes #154


lionkor (approved):
looks good, thank you

PR #288, opened 2024-02-12
Fix unicycle counting as car

### LINKED ISSUE #246: Unicycle is counted as a vehicle if you do not have a vehicle spawned
I have found that when in a session, if you spawn a unicycle and not a vehicle then when you go to spawn a vehicle after the server will not allow you.

If you remove your unicycle first and then spawn a vehicle it will allow you.

It looks as though the server counts your unicycle on some occasions as if you spawn the unicycle once you have a vehicle in a 1 vehicle per player server it will spawn as expected.

Does this make sense?

sla-ppy:
Fixes #246.

### DIFF
--- a/src/Client.cpp
+++ b/src/Client.cpp
@@ -120,13 +120,13 @@ void TClient::SetCarData(int Ident, const std::string& Data) {
     DeleteCar(Ident);
 }
 
-// TODO: count unicycles and vehicles different #246    sla-ppy
-// PROPOSAL: separate unicycles completely from vehicles???
 int TClient::GetCarCount() const {
+    // mVechileData holds both unicycle and cars which both count towards the maximum car count
+    // spawning a unicycle meant reaching the max, hence being unable to spawn car. this dirty fixes the problem for now.
     std::unique_lock lock(mVehicleDataMutex);
     for (auto& v : mVehicleData) {
         if (v.ID() == mUnicycleID) {
-            return int(mVehicleData.size() - 1); // dont keep track of unicycles
+            return int(mVehicleData.size() - 1);
         }
     }
     return int(mVehicleData.size());
--- a/src/Client.cpp
+++ b/src/Client.cpp
@@ -120,7 +120,15 @@ void TClient::SetCarData(int Ident, const std::string& Data) {
     DeleteCar(Ident);
 }
 
+// TODO: count unicycles and vehicles different #246    sla-ppy
+// PROPOSAL: separate unicycles completely from vehicles???
 int TClient::GetCarCount() const {
+    std::unique_lock lock(mVehicleDataMutex);
+    for (auto& v : mVehicleData) {
+        if (v.ID() == mUnicycleID) {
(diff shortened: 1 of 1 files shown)


sla-ppy:
This is a super dirty fix to the issue.
We essentially duct tape a problem which should definitely be revisited later on!

The reason for this dirty fix is because mVehichleData contains both unicycle and vehicle data, and I am unsure if we want it to stay like that.

I propose that in the new protocol we treat unicycle and vehicle data differently because there are questions to be answered like:
- Do we want to allow players to spawn multiple unicycles (as controllable props) or do we want to have a single one representing and acting as the player itself?
- Does it truly make sense to treat unicycles and vechicles differently? 

Benefit of the proposal:
- Counting the unicycle and vehicle counts separately could be toyed around with further.
- Future wise the unicycle could have multiple models for players to have fun with.

sla-ppy:
Please review my code: @lionkor.

src/Client.cpp
```diff
@@ -120,7 +120,15 @@ void TClient::SetCarData(int Ident, const std::string& Data) {
     DeleteCar(Ident);
 }
 
+// TODO: count unicycles and vehicles different #246    sla-ppy
+// PROPOSAL: separate unicycles completely from vehicles???
```
lionkor: please remove these comments lines 123 and 124

src/Client.cpp
```diff
@@ -120,7 +120,15 @@ void TClient::SetCarData(int Ident, const std::string& Data) {
...
 int TClient::GetCarCount() const {
+    std::unique_lock lock(mVehicleDataMutex);
+    for (auto& v : mVehicleData) {
```
lionkor, line 127: Please put a short(ish) comment here explaining the *reasoning* for this solution: For example, it should outline what this fixes, and why this is the fix that was chosen ("it works lol").
You can then remove the comment on L129.

lionkor (changes requested):
Thanks, logic looks sound and we obv saw it work.

Starystars67:
Could I further this topic by suggesting that instead a configuration option be added to specify vehicle names that can be added so they bypass the vehicle limit.

This would be for things like the unicycle as default but can then be expanded to items like cones, bollards, concrete barriers or even pucks.

lionkor:
> Could I further this topic by suggesting that instead a configuration option be added to specify vehicle names that can be added so they bypass the vehicle limit.
> 
> This would be for things like the unicycle as default but can then be expanded to items like cones, bollards, concrete barriers or even pucks.

I would ask you to make a new feature issue for that, because I'd also like to specify that properly before building it, and I wouldn't want to tack it onto this bug as I want this in the next release.

lionkor:
@sla-ppy btw in `documentate unicycle dirty fix` the correct form is `document` ;)

lionkor (approved):
looks good

sla-ppy:
> @sla-ppy btw in `documentate unicycle dirty fix` the correct form is `document` ;)

catty

PR #289, opened 2024-02-12
Fix timeout for synced players

### LINKED ISSUE #275: Reduce timeout for kicking players
Right now, the timeout for players is 20 minutes. This is unreasonable, and was only done because mod download pauses pings. Instead, we should check that the client is synced, and only kick people who are synced after a lower limit. We can keep the higher limit for unsynced players, of course.

sla-ppy:
Fix: #275.

While I was at it, I also refactored the debug output so its more appealing to read.

I dont see further ways to improve on this currently.

### DIFF
--- a/src/TPPSMonitor.cpp
+++ b/src/TPPSMonitor.cpp
@@ -65,10 +65,13 @@ void TPPSMonitor::operator()() {
                 V += c->GetCarCount();
             }
             // kick on "no ping"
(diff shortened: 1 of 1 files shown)


sla-ppy:
Please review my code: @lionkor.

lionkor (commented):
needs testing

lionkor:
@lionkor rebase this

PR #301, opened 2024-03-29
add a simple DOS protection system

Igor20264:
1) When a user connects to the server, I get his IP
2) When sending 'C', I check:
- if the ip is in blocked_ips.txt There is. I'm cutting the connection.
- if the user sent connection requests earlier (5 seconds) and their number within 5 seconds is more than 5, I write the ip to the file blocked_ips.txt and I'm disconnecting the connection.

The reaction time to the DOS ranges from 10 milliseconds to 1 second. P.S Very much depends on the strength of DOS. It should probably save you from DDOS attacks.

### DIFF
--- a/include/TNetwork.h
+++ b/include/TNetwork.h
@@ -27,14 +27,6 @@
 
 struct TConnection;
 
-class WatchingConnecting{
-public:
-    bool IsConnectionAllowed(const std::string& clientAddress);
-private:
-    void BlockIP(const std::string& clientAddress);
-    bool IsIPBlocked(const std::string& clientAddress);
-};
-
 class TNetwork {
 public:
     TNetwork(TServer& Server, TPPSMonitor& PPSMonitor, TResourceManager& ResourceManager);
--- a/include/TNetwork.h
+++ b/include/TNetwork.h
@@ -27,6 +27,14 @@
 
 struct TConnection;
 
+class WatchingConnecting{
+public:
+    bool IsConnectionAllowed(const std::string& clientAddress);
+private:
+    void BlockIP(const std::string& clientAddress);
+    bool IsIPBlocked(const std::string& clientAddress);
+};
+
 class TNetwork {
 public:
     TNetwork(TServer& Server, TPPSMonitor& PPSMonitor, TResourceManager& ResourceManager);
@@ -48,7 +56,6 @@ public:
 private:
     void UDPServerMain();
     void TCPServerMain();
-
     TServer& mServer;
     TPPSMonitor& mPPSMonitor;
     ip::udp::socket mUDPSock;
--- a/src/TNetwork.cpp
+++ b/src/TNetwork.cpp
@@ -20,6 +20,7 @@
 #include "Client.h"
 #include "Common.h"
 #include "LuaAPI.h"
+#include "RateLimiter.h"
 #include "TLuaEngine.h"
 #include "nlohmann/json.hpp"
 #include <CustomAssert.h>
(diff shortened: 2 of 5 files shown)


lionkor:
Thank you very much, I'll take a look at this soon :)

include/TNetwork.h
```diff
@@ -27,6 +27,14 @@
...
 struct TConnection;
 
+class WatchingConnecting{
+public:
+    bool IsConnectionAllowed(const std::string& clientAddress);
+private:
+    void BlockIP(const std::string& clientAddress);
+    bool IsIPBlocked(const std::string& clientAddress);
+};
+
```
lionkor: Please rename to something like `RateLimiter` and move it to its own file (`RateLimiter.cpp` and `RateLimiter.h`, for example).

src/TNetwork.cpp
```diff
@@ -31,6 +31,63 @@
...
+#include <unordered_map>
+#include <fstream>
+#include <chrono>
+#include <mutex>
+
+std::unordered_map<std::string, std::vector<std::chrono::time_point<std::chrono::high_resolution_clock>>> connectionAttempts;
+std::mutex connectionAttemptsMutex;
+
+bool WatchingConnecting::IsConnectionAllowed(const std::string& clientAddress) {
+        // we check if there is an IP in the blocked list
```
lionkor: please run `clang-format` on all files you modified, for example `clang-format -i --style=file src/*.cpp include/*.h`. This will fix all the indentation and other tiny things :)

src/TNetwork.cpp
```diff
@@ -31,6 +31,63 @@
...
 typedef boost::asio::detail::socket_option::integer<SOL_SOCKET, SO_RCVTIMEO> rcv_timeout_option;
 
+#include <iostream>
+#include <unordered_map>
+#include <fstream>
+#include <chrono>
+#include <mutex>
+
+std::unordered_map<std::string, std::vector<std::chrono::time_point<std::chrono::high_resolution_clock>>> connectionAttempts;
+std::mutex connectionAttemptsMutex;
```
lionkor: These should be stored in a class, I'd prefer as few globals as possible. It's much easier to test if they're not globals.

src/TNetwork.cpp
```diff
@@ -31,6 +31,63 @@
...
+        // we check if there is an IP in the blocked list
+        if (WatchingConnecting::IsIPBlocked(clientAddress)) {
+            return false;
+        }
+        std::lock_guard<std::mutex> lock(connectionAttemptsMutex);
+        auto currentTime = std::chrono::high_resolution_clock::now();
+        auto& violations = connectionAttempts[clientAddress];
+
+        // Deleting old violations (older than 5 seconds)
+        violations.erase(std::remove_if(violations.begin(), violations.end(),
```
lionkor: You could use [`erase_if`](https://en.cppreference.com/w/cpp/container/vector/erase2), but that's just a style thing

src/TNetwork.cpp
```diff
@@ -31,6 +31,63 @@
...
+                                 return std::chrono::duration_cast<std::chrono::seconds>(currentTime - timestamp).count() > 5;
+                             }), violations.end());
+
+        // Adding the current violation
+        violations.push_back(currentTime);
+
+        // We check the number of violations
+        if (violations.size() >= 4) {
+            WatchingConnecting::BlockIP(clientAddress);
+            beammp_errorf("[DOS] Blocked IP: {}", clientAddress);
```
lionkor:
```suggestion
            beammp_errorf("[DoS Protection] Blocked IP: {}", clientAddress);
```

src/TNetwork.cpp
```diff
@@ -31,6 +31,63 @@
...
+        std::ifstream blockFile("blocked_ips.txt");
+        std::unordered_set<std::string> blockedIPs;
+
+        if (blockFile.is_open()) {
+            std::string line;
+            while (std::getline(blockFile, line)) {
+                blockedIPs.insert(line);
+            }
+        }
+        return blockedIPs.find(clientAddress) != blockedIPs.end();
```
lionkor: You could use [`contains`](https://en.cppreference.com/w/cpp/container/unordered_map/contains) since this codebase is C++20, but you don't need to.

src/TNetwork.cpp
```diff
@@ -196,10 +253,16 @@ void TNetwork::Identify(TConnection&& RawConnection) {
...
     std::shared_ptr<TClient> Client { nullptr };
+    WatchingConnecting connectionManager;
     try {
-        if (Code == 'C') {
-            Client = Authentication(std::move(RawConnection));
+        if (Code == 'C') { 
+             if (connectionManager.IsConnectionAllowed(clientAddress)) {
+                Client = Authentication(std::move(RawConnection));
+             } else {
+                RawConnection.Socket.shutdown(socket_base::shutdown_both, ec);
```
lionkor, line 210: Could we maybe send a message here, and then close? It will be non-trivial to debug if this happens to someone by accident, they will think BeamMP is unstable.

lionkor (changes requested):
Thank you very much, nice work. Please make the changes I requested, if you can. Let me know if you have any questions :)

Igor20264:
Thanks for the criticism, I will try to correct these shortcomings in the near future

lionkor:
Hi @Igor20264, what's the status on this? I can make these changes, if you don't have time

Igor20264:
I completely forgot about this thread, sorry to keep you waiting. Unfortunately, due to problems with my studies, I will not be able to fix the code 😔. I'm sorry again.

lionkor:
@Igor20264 no worries! Thank you very much for your contribution. I'll make these fixes, and then merge it.

Don't worry, and good luck with your studies!

sla-ppy:
@lionkor I'd like to be assigned to make the changes suggested so this feature can be pushed out unless you'd like to do this yourself?

Good job on the implementation @Igor20264!

lionkor:
sure, i'd still like to make some changes potentially but only minor stuff (the constants), feel free.

src/TNetwork.cpp
sla-ppy: Every change suggested prior to this comment has been worked out so far.

I will commit when all of these suggestions have been implemented first, then we can think about what else we would like to further improve and act accordingly.

sla-ppy:
Oh yea, gotta update on the situation. So I managed to delete all the changes I made locally.
Since then I've moved to laptop-only and have to redo it all over again basically, working it rn.

PR #320, opened 2024-05-11
Use hard disconnect instead of ClientKick in timeout

lionkor:


OfficialLambdax (approved):
Approving the case that this is supposed to troubleshoot. Works in that regard in all test case. A lua hangup may still appear, but the server is able to deal with it in time and resume the hungup state

Issue #206, opened 2023-12-11
Compile Errors on macOS


max5800:
**General information**
OS:
`macOS 14.1 (23B74) - Intel`
BeamMP-Server Version:
`3.1.3`

**Describe the bug**
Cmake Build runs into errors. 

**To Reproduce**
1. 'brew install lua@5.3'
2. Have CommandLine-Tools installed
3. Run compile commands as documented in the readme. 
4. Get the following output:
https://pastebin.com/5GeQ4f3d 

**Expected behavior**
Get a Successful build.

**Additional context**
Could be an error on level 8, cause I have no idea what I'm doing.

lionkor:
Hi, this is a bug in https://github.com/lionkor/commandline, I will fix it there and upstream the fix into here.

Thanks for reporting this. Would you be willing to help me test it once fixed?

max5800:
Heyho,
thanks for the quick reply! I'm happy to help. 
Just keep me updated when and what to test. 
Danke!

max5800:
@lionkor 
I found two issues:
1. The commandline module wants to have explicitly c++11 which is now added to the cmake file. 
2.  Changed a line in TLuaEngine.cpp from ..nil to ..none 

Related [Pull Request 324](https://github.com/BeamMP/BeamMP-Server/pull/324)

max5800:
pull request merged, fix implemented, will be in next release. 
issue closed.

PR #324, opened 2024-05-18
Fix MacOS compile error

max5800:
Related [Issue 206](https://github.com/BeamMP/BeamMP-Server/issues/206)

### DIFF
--- a/src/TLuaEngine.cpp
+++ b/src/TLuaEngine.cpp
@@ -268,7 +268,7 @@ std::vector<std::string> TLuaEngine::StateThreadData::GetStateTableKeys(const st
 
     for (size_t i = 0; i < keys.size(); ++i) {
         auto obj = current.get<sol::object>(keys.at(i));
-        if (obj.get_type() == sol::type::none) {
+        if (obj.get_type() == sol::type::lua_nil) {
             // error
             break;
         } else if (i == keys.size() - 1) {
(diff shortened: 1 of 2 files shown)


lionkor:
I approved the CI/CD run on this, if it runs through I'll review + merge :) Thank you very much!

src/TLuaEngine.cpp
```diff
@@ -268,7 +268,7 @@ std::vector<std::string> TLuaEngine::StateThreadData::GetStateTableKeys(const st
 
     for (size_t i = 0; i < keys.size(); ++i) {
         auto obj = current.get<sol::object>(keys.at(i));
-        if (obj.get_type() == sol::type::nil) {
+        if (obj.get_type() == sol::type::none) {
```
lionkor:
```suggestion
        if (obj.get_type() == sol::type::lua_nil) {
```

none is a different type (LUA_TNONE) signifying no type, whereas lua_nil is the proper way to do `nil`

lionkor (changes requested):

src/TLuaEngine.cpp
max5800: Alright, I will test that later when I'm back at home.

max5800: changed to "lua_nil"  -> got successfull compiles on macos and my debian container. :)

lionkor (approved):
lgtm thank you very much

max5800:
You are welcome, it was fun!
Would be my first merged pull request. 😄🎉

lionkor:
Thanks, will be released in the next update

PR #325, opened 2024-05-24
Add platform, lua, openssl version to `version` command, show lua version on startup

### LINKED ISSUE #300: Print lua version + implementation on startup
When running a server and writing plugins for it, I want to know which lua version is used. On linux I can `ldd`, but on Windows I'm lost. The server should print the lua version and any other info it has about it, on startup of the LuaEngine.

sla-ppy:
Closes #300

### DIFF
--- a/src/TConsole.cpp
+++ b/src/TConsole.cpp
@@ -26,7 +26,9 @@
 #include "TLuaEngine.h"
 
 #include <ctime>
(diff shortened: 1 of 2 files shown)


lionkor (approved):
Looks good, thank you!

Waiting for the CI/CD to run through and then merging it.

PR #329, opened 2024-05-25
Add an entry in serverconfig.toml for the time between update reminders

### LINKED ISSUE #321: Please reduce update reminder console spam
Whenever there's a new version, the server prints an update reminder every 30 seconds or so.  I believe there's a bug in src/THeartbeatThread.cpp:132:

        if (!Application::Settings.HideUpdateMessages && UpdateReminderCounter % 5) {

where the server is supposed to check for updates every 5th heartbeat but instead it's checking it on every HB _except_ the 5th ones (I might be wrong, I have very little experience with C++).  But even the (I assume) intended period of 2.5 minutes seems pretty spammy, especially for small servers which don't see much traffic, where it would still outnumber all other ouput by an extreme margin.

Is it possible to reduce the frequency of update reminders?  Ideally having a parameter that would set the period manually in a configuration file (or at least a switch between minute-interval and day-interval reminders).

Setting the "ImScaredOfUpdates" flag in ServerConfig.toml gets rid of this issue, however it disables update reminders completely when the server is running.  I run my server continuously in a detached tmux instance so I don't restart it for weeks or months at a time (usually until there's a kernel update when I have to restart the system) so I do want to be reminded when there's a new version out. 

I'm using the Debian 12 build, but I don't think that's relevant here.

Acceptance criteria
---

- [ ] fix bug in update reminder to print every 5 heartbeats
- [ ] reduce update reminder to once every 3 or 6 hours on major versions (breaking changes) as we cant expect people to update immediately
- [ ]  Update reminder frequency setting with predefined values, like "hourly" and "daily" or something

SaltySnail:
fix #321


include/ChronoWrapper.h
```diff
@@ -0,0 +1,8 @@
+#pragma once
+
+#include <chrono>
+#include <string>
+
+namespace ChronoWrapper {
+    std::chrono::high_resolution_clock::duration TimeFromStringWithLiteral(const std::string time_str);
```
lionkor:
```suggestion
    std::chrono::high_resolution_clock::duration TimeFromStringWithLiteral(const std::string& time_str);
```

include/Common.h
```diff
@@ -151,7 +152,7 @@ class Application final {
     static inline std::mutex mShutdownHandlersMutex {};
     static inline std::deque<TShutdownHandler> mShutdownHandlers {};
 
-    static inline Version mVersion { 3, 5, 0 };
+    static inline Version mVersion { 4, 5, 0 };
```
lionkor:
```suggestion
    static inline Version mVersion { 3, 5, 0 };
```

src/ChronoWrapper.cpp
```diff
@@ -0,0 +1,36 @@
...
+    }
+    // else if (match.str(2) == "ms") {
+    //      return std::chrono::milliseconds(time_value);
+    // }
+    // else if (match.str(2) == "us") {
+    //      return std::chrono::microseconds(time_value);
+    // }
+    // else if (match.str(2) == "ns") {
+    //      return std::chrono::nanoseconds(time_value);
+    // }
```
lionkor: delete

lionkor (changes requested):

src/ChronoWrapper.cpp
SaltySnail: Removed the comments, mentioned in the requested change above, in: [e35bf4f](https://github.com/BeamMP/BeamMP-Server/pull/329/commits/e35bf4fe1520897b18820f11252ad6d557c859e8). Sorry for not giving +1 street cred

include/Common.h
```diff
@@ -83,6 +83,7 @@ class Application final {
         std::string HTTPServerIP { "127.0.0.1" };
         bool HTTPServerUseSSL { false };
         bool HideUpdateMessages { false };
+        std::string UpdateReminderTime { "3h" };
```
lionkor: Not sure if 3h is good, the default *should* be much quicker, like once every 30 mins(?)

lionkor (changes requested):

lionkor:
Also the maximum frequency is not possible, like 1s isn't possible, since it doesnt check for updates *that* much. But that is not a bug because this setting will only be used to turn the frequency *down*

include/Common.h
lionkor: Sorry, meant 30 seconds*

SaltySnail: fixed in [f3060f5](https://github.com/BeamMP/BeamMP-Server/pull/329/commits/f3060f5247b13325be5df87a18665bfc21ca7093).

lionkor (approved):
lgtm

PR #335, opened 2024-06-10
Add config setting to allow/deny guests

### LINKED ISSUE #247: Add "no guests" config setting
It would be nice to see a setting within the server configuration which disallows guest users from joining. 

The setting name would be like `authentication_required = true | false`

Ideally, this setting would also be communicated through heartbeat, so that end users would know that they would be able to join the server or not through the server list.

Acceptance Criteria
---
- [ ] Add "no guests" config setting
- [ ] Add equivalent ENV setting
- [ ] Send via heartbeat

SaltySnail:
fix #247

### DIFF
--- a/src/TConfig.cpp
+++ b/src/TConfig.cpp
@@ -52,7 +52,7 @@ static constexpr std::string_view EnvStrAuthKey = "BEAMMP_AUTH_KEY";
 static constexpr std::string_view StrLogChat = "LogChat";
 static constexpr std::string_view EnvStrLogChat = "BEAMMP_LOG_CHAT";
 static constexpr std::string_view StrAllowGuests = "AllowGuests";
-static constexpr std::string_view EnvStrAllowGuests = "BEAMMP_NO_GUESTS";
+static constexpr std::string_view EnvStrAllowGuests = "BEAMMP_ALLOW_GUESTS";
 static constexpr std::string_view StrPassword = "Password";
 
 // Misc
--- a/src/TConfig.cpp
+++ b/src/TConfig.cpp
@@ -51,6 +51,8 @@ static constexpr std::string_view StrAuthKey = "AuthKey";
 static constexpr std::string_view EnvStrAuthKey = "BEAMMP_AUTH_KEY";
 static constexpr std::string_view StrLogChat = "LogChat";
 static constexpr std::string_view EnvStrLogChat = "BEAMMP_LOG_CHAT";
+static constexpr std::string_view StrAllowGuests = "AllowGuests";
+static constexpr std::string_view EnvStrAllowGuests = "BEAMMP_NO_GUESTS";
 static constexpr std::string_view StrPassword = "Password";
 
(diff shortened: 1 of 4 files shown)


src/TConfig.cpp
```diff
@@ -51,6 +51,8 @@ static constexpr std::string_view StrAuthKey = "AuthKey";
 static constexpr std::string_view EnvStrAuthKey = "BEAMMP_AUTH_KEY";
 static constexpr std::string_view StrLogChat = "LogChat";
 static constexpr std::string_view EnvStrLogChat = "BEAMMP_LOG_CHAT";
+static constexpr std::string_view StrAllowGuests = "AllowGuests";
+static constexpr std::string_view EnvStrAllowGuests = "BEAMMP_NO_GUESTS";
```
lionkor: `BEAMMP_ALLOW_GUESTS` please to make it easy to guess the ENV

src/TNetwork.cpp
```diff
@@ -389,6 +389,12 @@ std::shared_ptr<TClient> TNetwork::Authentication(TConnection&& RawConnection) {
             return false;
         });
 
+    beammp_debugf("\t NotAllowedWithReason: {} \t AllowGuests: {} \t IsGuest: {}", NotAllowedWithReason, Application::Settings.AllowGuests, Client->IsGuest());
```
lionkor: is this debug print necessary? we could/should remove it or make it a `beammp_tracef` so it only shows up in debug builds with debug mode on

src/TNetwork.cpp
```diff
@@ -389,6 +389,12 @@ std::shared_ptr<TClient> TNetwork::Authentication(TConnection&& RawConnection) {
...
+    if (!NotAllowedWithReason && !Application::Settings.AllowGuests && Client->IsGuest()) { //!NotAllowedWithReason because this message has the lowest priority
+        NotAllowedWithReason = true;
+        Reason = "No guests are allowed on this server! To join, sign up at; forum.beammp.com.";
```
lionkor:
```suggestion
        Reason = "No guests are allowed on this server! To join, sign up at: forum.beammp.com.";
```

lionkor (changes requested):
Looks great, good solution there with the low prio deny of the auth, very nice. A few small things to fix ;)

src/TConfig.cpp
SaltySnail: fixed with [6e590ff](https://github.com/BeamMP/BeamMP-Server/pull/335/commits/6e590ff18ad3ea0b09649cdd4652f83003fe70d0)

src/TNetwork.cpp
SaltySnail: fixed with [e94cfd6](https://github.com/BeamMP/BeamMP-Server/pull/335/commits/e94cfd641d658688335d4789ee42195d440253d3)

lionkor (approved):
lgtm

PR #336, opened 2024-06-13
Fix disconnect not calling onVehicleDeleted

SaltySnail:
OnDisconnect sent a packet to the client, which was already disconnected.

### DIFF
--- a/src/TServer.cpp
+++ b/src/TServer.cpp
@@ -137,7 +137,6 @@ void TServer::RemoveClient(const std::weak_ptr<TClient>& WeakClientPtr) {
     beammp_assert(LockedClientPtr != nullptr);
     TClient& Client = *LockedClientPtr;
     beammp_debug("removing client " + Client.GetName() + " (" + std::to_string(ClientCount()) + ")");
-    // TODO: Send delete packets for all cars
     Client.ClearCars();
     WriteLock Lock(mClientsMutex);
     mClients.erase(WeakClientPtr.lock());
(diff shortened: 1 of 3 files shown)


lionkor:
How sure are you that this is the only place where we delete a vehicle? I recall there being more places where we *technically* delete a vehicle but dont do it properly, but im not sure. I'll take a look when i get time regardless

lionkor:
\+ thank you for taking on this task <3

src/TServer.cpp
```diff
@@ -333,6 +333,7 @@ void TServer::ParseVehicle(TClient& c, const std::string& Pckt, TNetwork& Networ
                     // TODO: handle
                 }
                 std::string Destroy = "Od:" + std::to_string(c.GetID()) + "-" + std::to_string(CarID);
+                //LuaAPI::MP::Engine->ReportErrors(LuaAPI::MP::Engine->TriggerEvent("onVehicleDeleted", "", c.GetID(), CarID)); //uncomment if this packet is sent to everyone instead of to the client that spawned too many vehicles
```
lionkor: remove pls

lionkor (changes requested):

src/TServer.cpp
lionkor: or... should it be there?

SaltySnail: It needed to be there

lionkor (commented):
Looks good, will need to test it which can take some time. Thank you!

PR #342, opened 2024-06-25
Fix little issues

lionkor:


src/Common.cpp
```diff
@@ -386,3 +387,51 @@ void SplitString(const std::string& str, const char delim, std::vector<std::stri
     }
 }
 
+std::vector<uint8_t> DeComp(std::span<const uint8_t> input) {
+    beammp_debugf("got {} bytes of input data", input.size());
+
+    std::vector<uint8_t> output_buffer(std::min<size_t>(input.size() * 5, 15 * 1024 * 1024));
```
jimkoen: What about a short comment detailing the constants in the vector constructor? As someone who hasn't touched this code it isn't immediately apparent what these do.

src/Common.cpp
```diff
@@ -386,3 +387,51 @@ void SplitString(const std::string& str, const char delim, std::vector<std::stri
...
+    uLongf output_size = output_buffer.size();
+
+    while (true) {
+        int res = uncompress(
+            reinterpret_cast<Bytef*>(output_buffer.data()),
+            &output_size,
+            reinterpret_cast<const Bytef*>(input.data()),
+            static_cast<uLongf>(input.size()));
+        if (res == Z_BUF_ERROR) {
+            if (output_buffer.size() > 30 * 1024 * 1024) {
```
jimkoen: Or just a short description why this is needed.

jimkoen (commented):
Wasn't so sure about the magic constants, but I need to clarify that I never touch this code.

Other than that, lgtm

jimkoen (approved):

src/Common.cpp
lionkor: good point, thanky ou

src/Common.cpp
lionkor: good point here as well, thanks

src/Common.cpp
```diff
@@ -386,3 +387,63 @@ void SplitString(const std::string& str, const char delim, std::vector<std::stri
...
+            beammp_warnf("zlib uncompress() failed, trying with a larger buffer size of {}", output_buffer.size());
+            output_size = output_buffer.size();
+        } else if (res != Z_OK) {
+            beammp_error("zlib uncompress() failed: " + std::to_string(res));
+            throw std::runtime_error("zlib uncompress() failed");
+        } else if (res == Z_OK) {
+            break;
+        }
+    }
+    output_buffer.resize(output_size);
```
jimkoen, line 429: Isn't this superfluous given line 421?

jimkoen (dismissed):
lgtm

PR #349, opened 2024-07-13
Add IPv6 support

lionkor:
Adds IPv6 support.

FreeBSD users beware: From this point forward, you will have to set `sysctl net.inet6.ip6.v6only=0` so that the BeamMP-Server continues to work. The server also prints this information.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -100,7 +100,7 @@ set(PRJ_LIBRARIES
     commandline_static
     toml11::toml11
     rapidjson
-    sol2
+    sol2::sol2
     httplib::httplib
     libzip::zip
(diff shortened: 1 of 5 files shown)


jimkoen (approved):
compiles and runs on FreeBSD

lgtm

edit: revoking my approval bc this breaks ipv4

SaltySnail (approved):
Looks good, works on Windows

lionkor:
FreeBSD users beware: From this point forward, you will have to set sysctl net.inet6.ip6.v6only=0 so that the BeamMP-Server continues to work. The server also prints this information.

SaltySnail:
LGTM, no functionality changed since I last reviewed, so I will go ahead and merge

PR #352, opened 2024-07-27
Move toml11 out of vcpkg since the toml11 authors broke it

lionkor:


WiserTixx:
Am I missing something, or has the vcpkg file not been changed?

PR #366, opened 2024-09-19
Add Lua function to get a player's role

WiserTixx:
Adds `MP.GetPlayerRole(player_id)` to the Lua API to get a player's role ("USER", "EA", "MDEV", "STAFF", "ET") by their player id.
Currently you can only get someone's role in onPlayerAuth from the parameters and in onVehicleSpawned and onVehicleEdited from the packet data, but not in onPlayerJoin for example without storing it.

### DIFF
--- a/src/TLuaEngine.cpp
+++ b/src/TLuaEngine.cpp
@@ -597,10 +597,11 @@ sol::table TLuaEngine::StateThreadData::Lua_GetPlayerIdentifiers(int ID) {
 
 std::string TLuaEngine::StateThreadData::Lua_GetPlayerRole(int ID) {
     auto MaybeClient = GetClient(mEngine->Server(), ID);
-    if (MaybeClient && !MaybeClient.value().expired())
+    if (MaybeClient) {
         return MaybeClient.value().lock()->GetRoles();
-    else
+    } else {
         return "";
+    }
 }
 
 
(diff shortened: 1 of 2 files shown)


src/TLuaEngine.cpp
```diff
@@ -595,6 +595,15 @@ sol::table TLuaEngine::StateThreadData::Lua_GetPlayerIdentifiers(int ID) {
     }
 }
 
+std::string TLuaEngine::StateThreadData::Lua_GetPlayerRole(int ID) {
+    auto MaybeClient = GetClient(mEngine->Server(), ID);
+    if (MaybeClient && !MaybeClient.value().expired())
+        return MaybeClient.value().lock()->GetRoles();
+    else
+        return "";
```
lionkor: Please use braces

lionkor: and remove the expiry check, it's been done in the past but doesn't actually do anything :)

lionkor (changes requested):

WiserTixx:
Requested changes have been made.

src/TLuaEngine.cpp
```diff
@@ -595,6 +595,16 @@ sol::table TLuaEngine::StateThreadData::Lua_GetPlayerIdentifiers(int ID) {
     }
 }
 
+std::string TLuaEngine::StateThreadData::Lua_GetPlayerRole(int ID) {
+    auto MaybeClient = GetClient(mEngine->Server(), ID);
+    if (MaybeClient) {
+        return MaybeClient.value().lock()->GetRoles();
+    } else {
+        return "";
```
lionkor: Could we please return `nil` instead?

PR #385, opened 2024-11-01
Fix backend heartbeat

WiserTixx:


lionkor:
Ew.

lionkor:
Thanks for fixing it <3 but wow.

0rz3:
> Ew.

gv

PR #413, opened 2025-01-16
Fix crash

WiserTixx:
This PR fixes the segfault caused by rapidjson on the first heartbeat, and the memory leak that happens during mod hashing.

---


SaltySnail (approved):
lgtm

PR #417, opened 2025-01-19
Force IPv4 for backend requests

WiserTixx:
Force the use of ipv4 for backend requests because the launcher doesn't support ipv6 yet

---


WiserTixx:
This has been tested and seems to work

snepsnepsnep (approved):

PR #419, opened 2025-01-25
Added check for Stateview, added socket cancel to finish all connect,…

SaltySnail:
Fix for some race condition related crashes of the server. 
---


lionkor:
Question from the sidelines; boost docs [say that close() already cancels](https://www.boost.org/doc/libs/1_87_0/doc/html/boost_asio/reference/basic_stream_socket/cancel/overload1.html)

> Use the close() function to simultaneously cancel the outstanding operations and close the socket. 

What does this do internally, as we already close() right after? 

Wondering since I use this API in other projects lol

PR #425, opened 2025-03-12
Add custom IP bind option

### LINKED ISSUE #57: Allow Custom IP to bind to
Currently you can specify a custom IP via the first argument to the server. It's then subsequently ignored by the backend. 

The entire thing needs to be an optional config setting, and then be bound to with bind() so that we can ensure the IP is actually owned by the server.

### LINKED ISSUE #173: [Bug] Out-of-range port does not generate an error
OS (windows, linux, ...): Windows 10
BeamMP-Server Version: 3.1.1

**Describe the bug**
You can set a port in the serverConfig.toml that is above the allowed IPv4 portrange of 65535.
The server binary will happily accept that port without an error, but of course theres no way to connect to that port

**To Reproduce**
1. Enter a port number higher than 65535 in the serverConfig.toml
2. Start the server binary
3. It will report that the server is successfully running on the chosen port
4. But theres no way to connect to that port

**Expected behavior**
A .toml file will accept any input of course, but at least the server binary should throw an error when the port is out-of-range
Adding to that, too low of a port (<1024) should also throw an error.

**Additional context**
image

WiserTixx:
This PR adds an option in the server config to bind to a custom IP.

- [x] Review config comment and make sure that it is not confused with port forwarding
- [x] Validate IP addresses

---


src/TConfig.cpp
```diff
@@ -138,6 +140,8 @@ void TConfig::FlushToFile() {
     data["General"][StrInformationPacket.data()] = Application::Settings.getAsBool(Settings::Key::General_InformationPacket);
     data["General"][StrAllowGuests.data()] = Application::Settings.getAsBool(Settings::Key::General_AllowGuests);
     SetComment(data["General"][StrAllowGuests.data()].comments(), " Whether to allow guests");
+    data["General"][StrIP.data()] = Application::Settings.getAsString(Settings::Key::General_IP);
+    SetComment(data["General"][StrIP.data()].comments(), " The IP address to bind the server to, this is NOT related to your public IP.");
```
O1LER:
```suggestion
    SetComment(data["General"][StrIP.data()].comments(), "The IP address to bind the server to, this is NOT related to your public IP. Can be used if your machine has multiple network interfaces");
```

src/TNetwork.cpp
```diff
@@ -209,6 +222,7 @@ void TNetwork::TCPServerMain() {
         Application::GracefullyShutdown();
     }
     Application::SetSubsystemStatus("TCPNetwork", Application::Status::Good);
+    beammp_infof("Listening on {0} port {1}", ListenEp.address().to_string(), (int)ListenEp.port());
```
SaltySnail: Would've recommended a c++ style cast, but can't see a reason that uint16_t casting to an int would break. Although it is easier to find a static_cast with ctrl+f

SaltySnail (approved):
lgtm

pedrotski:
Only just noticed this update. 

Can this please be made to work the same way as PROVIDER_PORT_ENV does. As of right now, it's not going to work on Pterodatyl or WISP out of the box without additional configuration.

WiserTixx:
> Can this please be made to work the same way as PROVIDER_PORT_ENV does.

Sure thing, see  #432.

PR #426, opened 2025-03-24
Self check functionality

### LINKED ISSUE #210: [Feature Request] Connection test between  server and backend
The serverlist conatins servers that are not port forwarded properly, or have a similar issue that impedes the ability to connect to them.

On server startup, the server checks in on the backend and informs the owner by stating
 ```
 [17/12/23 12:19:13] [INFO] Resumed authenticated session! Success, Your server listing has been updated.
[17/12/23 12:19:13] [INFO] ALL SYSTEMS STARTED SUCCESSFULLY, EVERYTHING IS OKAY
```

But in reality this check was only for the connection server-backend. It does not mean that users connecting from the serverlist are actually able to connect.

Upon receiving information of a new/restarted server, the backend should do a sanity check on the posted port to determine if actual users are able to connect to that port.
If thats the case, the server console can state the `ALL SYSTEMS STARTED SUCCESSFULLY, EVERYTHING IS OKAY`
If the port is closed, or the server is otherwise not reachable via the posted port, then in fact not everything is okay. Server owner should be notified of that in the server console

The server could do a connection check to its port via a third party service, like a portchecker website. If thats available for automatic use

### LINKED ISSUE #278: Built in port tester
A service that would check if the server is accessible from the outside on startup, ideally for both TCP&UDP and print the result in the console. It would save a lot of time with troubleshooting port forwarding and users wouldn't have to use a third party port checking tool.

A backend api that would be called by the server on startup. It would have to do more than a simple TCP connection to check for UDP connectivity, (ie attempt to send something and check for response).

Not doing that and having users continue to use third party tools.

**Additional context**
I am aware that the UDP part is not possible in the current release, but just something to think about for future releases.

Additionally if possible having a built in detection for NAT loopback support, if it's not supported users won't be able to join their server from the server list even if they've forwarded everything correctly, which obviously leads to frustration and less than optimal experience.
(example behaviour: if the open port test passes, attempt connection to itself via public IP and see if it works, if not print out a message telling the user to use local IP address when connecting from the game).

WiserTixx:
This PR adds a new console command (`nettest`) that sends a request to the server check api in order to test connectivity via the server's public ip (serverlist entry).

- [x] https://github.com/BeamMP/ServerCheck/pull/2

---

### DIFF
--- a/include/Common.h
+++ b/include/Common.h
@@ -86,6 +86,8 @@ public:
         };
     }
 
+    static std::string GetServerCheckUrl() { return "https://check.beammp.com"; }
+
     static std::string GetBackendUrlForAuth() { return "https://auth.beammp.com"; }
     static std::string GetBackendUrlForSocketIO() { return "https://backend.beammp.com"; }
     static void CheckForUpdates();
--- a/include/TConsole.h
+++ b/include/TConsole.h
@@ -61,6 +61,7 @@ private:
     void Command_Version(const std::string& cmd, const std::vector<std::string>& args);
     void Command_ProtectMod(const std::string& cmd, const std::vector<std::string>& args);
     void Command_ReloadMods(const std::string& cmd, const std::vector<std::string>& args);
+    void Command_NetTest(const std::string& cmd, const std::vector<std::string>& args);
 
     void Command_Say(const std::string& FullCommand);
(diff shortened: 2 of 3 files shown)


O1LER:
Closes #278

SaltySnail (approved):
lgtm

PR #427, opened 2025-03-30
Implement Dialog packet and add MP.ConfirmationDialog

WiserTixx:
This PR implements a new lua function and packet used for sends dialogs to the client.

## Example:

https://github.com/user-attachments/assets/97bb5813-ea12-4b1d-a049-2f7ebf6b6da3

Example serverside code:
```lua
--MP.ConfirmationDialog(player_id: number, title: string, body: string, buttons: object, interaction_id: string, warning: boolean = false, reportToServer: boolean = true, reportToExtensions: boolean = true)

function onChatMessage(player_id, player_name, message)
    MP.ConfirmationDialog(player_id, "Warning", "Watch your tone buddy!!", 
        {
            {
                label = "OK",
                key = "dialogOK",
                isCancel = true
            }
        }, "interactionID", true)
end

MP.RegisterEvent("onChatMessage", "onChatMessage")


function dialogOK(player_id, interaction_id)
    MP.SendChatMessage(-1, MP.GetPlayerName(player_id) .. " clicked OK")
end

MP.RegisterEvent("dialogOK", "dialogOK")
```

### Details:
Each dialog can have multiple buttons, each button having it's own callback event (`key`).
Each dialog can also have one button with `isCancel` being true, settings this property to true causes the button's event to be called when the users pressed `esc` to exit out of the dialog. If a dialog is created without any button being the cancel button then the user will only be able to exit the dialog by restarting the session or pressing one of the buttons.

`interaction_id` will be sent as the event data with a button press event, to track from which dialog the button press came. As when multiple dialogs are opened they will stack and it will become difficult to track what button on which dialog was pressed without having multiple event handlers.

Waiting on https://github.com/BeamMP/BeamMP/pull/715 to be merged.

---

### DIFF
--- a/include/LuaAPI.h
+++ b/include/LuaAPI.h
@@ -36,7 +36,7 @@ namespace MP {
     std::pair<bool, std::string> DropPlayer(int ID, std::optional<std::string> MaybeReason);
     std::pair<bool, std::string> SendChatMessage(int ID, const std::string& Message);
     std::pair<bool, std::string> SendNotification(int ID, const std::string& Message, const std::string& Icon, const std::string& Category);
-    std::pair<bool, std::string> ConfirmationDialog(int ID, const std::string& Title, const std::string& Body, const sol::table& buttons, const std::string& InteractionID, const bool& warning = false);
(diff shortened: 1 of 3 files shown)


SaltySnail (approved):
Cool feature, lgtm. Be sure to add the example code to the docs so  people know how the format the packet

SaltySnail (approved):
lgtm

PR #428, opened 2025-04-06
Revert "support for nested lua handlers"

WiserTixx:
Reverts a PR that has been causing sol to crash.

---

### DIFF
--- a/src/TLuaEngine.cpp
+++ b/src/TLuaEngine.cpp
@@ -40,18 +40,6 @@ TLuaEngine* LuaAPI::MP::Engine;
 
 static sol::protected_function AddTraceback(sol::state_view StateView, sol::protected_function RawFn);
 
-static std::optional<sol::function> GetLuaHandler(sol::state_view StateView, const std::string Handler, const std::string EventName);
-
-static std::optional<sol::function> GetLuaHandler(sol::state_view StateView, const std::string Handler, const std::string EventName) {
-    auto Res = StateView.safe_script("return " + Handler, sol::script_pass_on_error);
-    if (!Res.valid()) {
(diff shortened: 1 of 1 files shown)


src/TLuaEngine.cpp
```diff
@@ -507,11 +495,9 @@ sol::table TLuaEngine::StateThreadData::Lua_TriggerGlobalEvent(const std::string
 
     sol::variadic_results LocalArgs = JsonStringToArray(Str);
     for (const auto& Handler : MyHandlers) {
-        auto Res = GetLuaHandler(mStateView, Handler, EventName);
-        if (Res.has_value()) {
-            sol::function Fn = Res.value();
-            Fn = AddTraceback(mStateView, Fn);
-
+        auto Fn = mStateView[Handler];
```
SaltySnail, line 498: Would this never segfault? Seems like there should be a check.

SaltySnail (changes requested):

src/TLuaEngine.cpp
WiserTixx: I believe the check on line 500 should do so.

SaltySnail: No I mean that mStateView[Handler] could segfault if Handler is out of bounds of the mStateView. But maybe it is ensured somewhere else that that doesn't happen.

SaltySnail (changes requested):

src/TLuaEngine.cpp
SaltySnail: Ooh yeah reading up on it it is handled in the sol::state_view::operator[].

SaltySnail (approved):
lgtm

PR #429, opened 2025-04-13
Make PluginMonitor not try to run non-lua files

WiserTixx:

### DIFF
--- a/src/TPluginMonitor.cpp
+++ b/src/TPluginMonitor.cpp
@@ -57,21 +57,26 @@ void TPluginMonitor::operator()() {
                     mFileTimes[Pair.first] = CurrentTime;
                     // grandparent of the path should be Resources/Server
                     if (fs::equivalent(fs::path(Pair.first).parent_path().parent_path(), mPath)) {
-                        beammp_infof("File \"{}\" changed, reloading", Pair.first);
-                        // is in root folder, so reload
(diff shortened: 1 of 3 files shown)


src/TPluginMonitor.cpp
```diff
@@ -57,21 +57,26 @@ void TPluginMonitor::operator()() {
...
-                        auto Contents = std::make_shared<std::string>();
-                        Contents->resize(Size);
-                        FileStream.read(Contents->data(), Contents->size());
-                        TLuaChunk Chunk(Contents, Pair.first, fs::path(Pair.first).parent_path().string());
-                        auto StateID = mEngine->GetStateIDForPlugin(fs::path(Pair.first).parent_path());
-                        auto Res = mEngine->EnqueueScript(StateID, Chunk);
-                        Res->WaitUntilReady();
-                        if (Res->Error) {
-                            beammp_lua_errorf("Error while hot-reloading \"{}\": {}", Pair.first, Res->ErrorMessage);
+                        if (fs::path(Pair.first).extension() == ".lua") {
```
OfficialLambdax: Question: Does the extension method return the extension as lowercase? from the std docs i cannot see that. Thinking is that eg .LUA will slip by this

SaltySnail (approved):
lgtm

PR #430, opened 2025-04-26
Client resource hash database and client resource protection

WiserTixx:
# Mod database
This PR adds a local database of mods, which is used to cache mod hashes and protection status.

## Mod hash caching
Mod hashes will now be cached based on last write date. This will speed up server startup because only the mods with changes will have to be hashed.

## Mod protection
You can now protect mods! This will allow you to host a server with copyrighted content without actually hosting the copyrighted content. Just run `protectmod <filename with .zip> <true/false>` in the console to protect a mod. Users that join a server with protected mods will have to obtain the file themselves and put it in their launcher's resources folder. The launcher will inform the user about this if the file is missing.

## Mod reloading
You can now reload client mods while the server is running by using `reloadmods` in the console. Keep in mind that this is mainly intended for development, therefore it will **not** force client to rejoin and neither will is hot-reload mods on the client.

---

### DIFF
--- a/src/TConsole.cpp
+++ b/src/TConsole.cpp
@@ -218,7 +218,7 @@ void TConsole::Command_Help(const std::string&, const std::vector<std::string>& 
         status                     how the server is doing and what it's up to
         clear                      clears the console window
         version                    displays the server version
-        protectmod <name> <value>  sets whether a mod is protected
+        protectmod <name> <value>  sets whether a mod is protected, value can be true or false
         reloadmods                 reloads all mods from the Resources Client folder)";
     Application::Console().WriteRaw("BeamMP-Server Console: " + std::string(sHelpString));
 }
--- a/src/TConsole.cpp
+++ b/src/TConsole.cpp
@@ -208,16 +208,18 @@ void TConsole::Command_Help(const std::string&, const std::vector<std::string>& 
     }
     static constexpr const char* sHelpString = R"(
     Commands:
-        help                    displays this help
-        exit                    shuts down the server
-        kick <name> [reason]    kicks specified player with an optional reason
-        list                    lists all players and info about them
-        say <message>           sends the message to all players in chat
-        lua [state id]          switches to lua, optionally into a specific state id's lua
-        settings [command]      sets or gets settings for the server, run `settings help` for more info
-        status                  how the server is doing and what it's up to
-        clear                   clears the console window
-        version                 displays the server version)";
+        help                       displays this help
+        exit                       shuts down the server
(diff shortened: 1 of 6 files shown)


src/TConsole.cpp
```diff
@@ -208,16 +208,18 @@ void TConsole::Command_Help(const std::string&, const std::vector<std::string>&
...
+        exit                       shuts down the server
+        kick <name> [reason]       kicks specified player with an optional reason
+        list                       lists all players and info about them
+        say <message>              sends the message to all players in chat
+        lua [state id]             switches to lua, optionally into a specific state id's lua
+        settings [command]         sets or gets settings for the server, run `settings help` for more info
+        status                     how the server is doing and what it's up to
+        clear                      clears the console window
+        version                    displays the server version
+        protectmod <name> <value>  sets whether a mod is protected
```
SaltySnail:
```suggestion
        protectmod <name> <value>  sets whether a mod is protected, value can be true or false
```

src/TNetwork.cpp
```diff
@@ -808,6 +808,15 @@ void TNetwork::SendFile(TClient& c, const std::string& UnsafeName) {
         return;
     }
     auto FileName = fs::path(UnsafeName).filename().string();
+
+    for (auto mod : mResourceManager.GetMods()) {
+        if (mod["file_name"].get<std::string>() == FileName && mod["protected"] == true) {
+            beammp_warn("Client tried to access protected file " + UnsafeName);
+            c.Disconnect("Mod is protected thus cannot be downloaded");
+            return;
+        }
```
SaltySnail, line 817:
```suggestion
        const std::string& ModFileName = mod["file_name"].get<std::string>()
        if (ModFileName == FileName && mod["protected"] == true) {
            beammp_warn("Client tried to access protected file " + UnsafeName);
            c.Disconnect("The mod " + ModeFileName + " is protected, add it manually to join.");
            return;
        }
```

Feel free to reject this change if it doesn't make sense, I think it should be clear for the user why they can't join

SaltySnail (changes requested):
Code looks good, just some suggestions on stuff the user is going to see

src/TNetwork.cpp
WiserTixx: The client will never see this message, as the launcher will check wether a mod is protected and then disconnect and show a more detailed message to the user if it's missing.
This is more a message for when a 3rd party or malicious client tries to get a protected mod.

SaltySnail (approved):
lgtm

PR #431, opened 2025-05-01
Prevent lua sending client events during downloading

### LINKED ISSUE #157: [Bug] Server events are sent to the client before they have "synced"
**Describe the bug**
The launcher interprets "E" packets as kick packets pre-sync causing the user to get kicked by events.

**Solution**
Don't send any events to the client until they have "synced".

WiserTixx:
This PR fixes an issue where players would get personal events during downloads, which would corrupt the download and block the user from being able to properly join the server.

---


SaltySnail (approved):
lgtm

PR #438, opened 2025-08-05
Show IPs in list console command

vadyankaww:
Added display of player's IP in list

---

### DIFF
--- a/src/TConsole.cpp
+++ b/src/TConsole.cpp
@@ -564,13 +564,14 @@ void TConsole::Command_List(const std::string&, const std::vector<std::string>& 
         Application::Console().WriteRaw("No players online.");
     } else {
         std::stringstream ss;
-        ss << std::left << std::setw(25) << "Name" << std::setw(6) << "ID" << std::setw(6) << "Cars" << std::endl;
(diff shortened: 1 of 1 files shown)


WiserTixx:
Hey, could you explain why you've made this change and what its use cases are?

vadyankaww:
> Hey, could you explain why you've made this change and what its use cases are?

This way it will be possible to identify twink accounts, which is quite useful on some servers.

vadyankaww:
and in general it is useful for many servers

WiserTixx:
> This way it will be possible to identify twink accounts, which is quite useful on some servers.

Couldn't this also be done automatically in a server script?

vadyankaww:
> Couldn't this also be done automatically in a server script?

It is possible, but why load the server with unnecessary code?

vadyankaww:
and agree, not everyone knows how to write a script, and not everyone needs it

WiserTixx:
This seems like unnecessary security risk. If this is released and people run the `list` they will suddenly get a list of people's ip addresses, which could lead to unexpected situations and possible data protection issues.
Identifying possible malicious accounts would be too specific to implement into the server, and not very convenient for the server operator. Ultimately a script should be used for this.

vadyankaww:
mb

PR #445, opened 2025-12-12
Fix build

alextwothousand:
Appears as if toml11 was causing issues with compilation.

I have decided to pull it via CMake FetchContent, as it will fetch & compile the library, as intended, by the author.

This has fixed the issue for me on Ubuntu, albeit via WSL2.

---


alextwothousand:
Could we re-run CI with the new workflow?

PR #448, opened 2025-12-24
add native color code support in console output

Kipstz:
This PR adds native support for color codes in console output, allowing plugin developers to easily colorize their server logs and messages using simple ^X codes.

  The Problem

  Currently, if developers want colored console output, they need to either:
  - Manually write ANSI escape sequences (e.g., \x1b[31m for red) which are hard to remember and verbose
  - Use external Lua libraries or workarounds

  The Solution

  This implementation introduces intuitive color codes inspired by classic game server conventions (similar to Cfx, FiveM, RedM):

  -- Before (raw ANSI - hard to read and write)
  print("\x1b[31mError: \x1b[0mSomething went wrong")

  -- After (simple and intuitive)
  print("^4Error: ^rSomething went wrong")

  Available Codes

  | Code | Effect            | Code | Effect    |
  |------|-------------------|------|-----------|
  | ^r   | Reset             | ^l   | Bold      |
  | ^o   | Italic            | ^n   | Underline |
  | ^m   | ~~Strikethrough~~ | ^p   | New line  |

  | Code | Color   | Code | Color         |
  |------|---------|------|---------------|
  | ^0   | Black   | ^8   | Dark Gray     |
  | ^1   | Blue    | ^9   | Light Blue    |
  | ^2   | Green   | ^a   | Light Green   |
  | ^3   | Cyan    | ^b   | Light Cyan    |
  | ^4   | Red     | ^c   | Light Red     |
  | ^5   | Magenta | ^d   | Light Magenta |
  | ^6   | Yellow  | ^e   | Light Yellow  |
  | ^7   | White   | ^f   | Bright White  |

  Examples

  -- Colored status messages
  print("^2[SUCCESS]^r Player connected")
  print("^4[ERROR]^r Connection failed")
  print("^6[WARNING]^r Server is full")
  image

  -- Formatted text
  print("^lBold^r and ^oitalic^r text")
  image

  -- Multiple colors
  print("^1Blue ^2Green ^4Red ^rNormal")
  image

  Why This Should Be Merged

  1. Intuitive: Two-character codes are easy to remember (^4 = red, ^2 = green)
  2. Familiar: Similar syntax to Quake, Minecraft, and other game servers - many developers already know it
  3. Non-breaking: Unknown codes (like ^z) are left unchanged, ensuring backward compatibility
  4. Lightweight: Simple string parsing with no external dependencies
  5. Tested: Includes unit tests for all color codes

### DIFF
--- a/src/TConsole.cpp
+++ b/src/TConsole.cpp
@@ -30,6 +30,7 @@
 #include <lua.hpp>
 #include <mutex>
 #include <openssl/opensslv.h>
+#include <regex>
 #include <sstream>
 #include <stdexcept>
 #include <unordered_map>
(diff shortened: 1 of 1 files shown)


src/TConsole.cpp
```diff
@@ -896,7 +953,7 @@ void TConsole::InitializeCommandline() {
 }
 
 void TConsole::Write(const std::string& str) {
-    auto ToWrite = GetDate() + str;
+    auto ToWrite = GetDate() + ConvertColorCodes(str);
```
WiserTixx, line 956: Could this be moved to a higher level function? I'd prefer this to be part of a `printColored` function (in-line with `print` and `printRaw` in lua, c++ code can just call this helper) instead of applying to everything going to the console.

WiserTixx (changes requested):

src/TConsole.cpp
Kipstz: That's my opinion, but I don't necessarily see the point of having several print functions. You might as well have one that handles most things. That's how it works right now: print(“\x1b[31mError: \x1b[0mSomething went wrong”). I just simplified it by doing this: print("^4Error: ^rSomething went wrong"). I think it would be overkill to have fun creating new print functions.

WiserTixx: I don't want server scripts to suddenly output unexpected text, however I suppose the original will still be present in the server log.
Either way, this should be something applied in the lua api handler, not for everything going to the terminal.

PR #451, opened 2025-12-24
refactor: optimize string operations and improve code clarity

Kipstz:
- Avoid redundant substr() calls in packet parsing hot-path (TServer.cpp)
  The previous code called substr(3) twice per packet, creating
  unnecessary temporary strings. Now stores the result once.

- Replace .size() == 0 with .empty() for idiomatic C++
  (TConsole.cpp, TLuaEngine.cpp)

### DIFF
--- a/src/TServer.cpp
+++ b/src/TServer.cpp
@@ -198,7 +198,8 @@ void TServer::GlobalParser(const std::weak_ptr<TClient>& Client, std::vector<uin
         int PID = -1;
         int VID = -1;
 
-        auto MaybePidVid = GetPidVid(StringPacket.substr(3).substr(0, StringPacket.substr(3).find(':', 1)));
+        auto pidVidPart = StringPacket.substr(3);
+        auto MaybePidVid = GetPidVid(pidVidPart.substr(0, pidVidPart.find(':')));
         if (MaybePidVid) {
             std::tie(PID, VID) = MaybePidVid.value();
         }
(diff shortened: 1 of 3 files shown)


src/TServer.cpp
```diff
@@ -289,7 +290,8 @@ void TServer::GlobalParser(const std::weak_ptr<TClient>& Client, std::vector<uin
         int PID = -1;
         int VID = -1;
 
-        auto MaybePidVid = GetPidVid(StringPacket.substr(3).substr(0, StringPacket.substr(3).find(':', 1)));
+        auto pidVidPartZ = StringPacket.substr(3);
```
WiserTixx: Why does the variable name end in a `Z`?

WiserTixx (changes requested):

src/TServer.cpp
Kipstz: Because I already use the variable without z at the top, so I added a z. I didn't have any other ideas.

WiserTixx: They're not in the same scope, though, so it wouldn't conflict.

Kipstz: I changed them, I modified them.

WiserTixx:
Is there any reason why you didn't use the pull request template?

Kipstz:
> Is there any reason why you didn't use the pull request template?

I didn't even notice there was a template.

WiserTixx (approved):

PR #452, opened 2025-12-24
feat: add onInitFinal event for post-initialization synchronization

### LINKED ISSUE #434: [Feature Request] onInitFinal - A event called when all lua states have been initialized
When Lua states want to communicate with each other then they have to wait for the other to exist. So if a script wants to eg register, request or send data during eg onInit
```lua
MP.TriggerGlobalEvent("myEvent", "hey im here")
```
then this event will only reach the target state when it has been initialized already.

For example imagine a single state that manages all server wide script settings. Now each script state has to request that data, but has to make sure the options state is already present. Currently there is no way to ensure a or all states are fully loaded.

An event like onInitFinal that is fired once all states have been fully loaded and executed with onInit would solve this.

Kipstz:
Adds a new 'onInitFinal' event that triggers after ALL Lua states have completed their 'onInit' handlers. This provides a reliable synchronization point for inter-state communication.

Use case: When multiple Lua states need to communicate (e.g., a centralized settings manager), they can now safely interact in onInitFinal knowing all states are fully initialized.

Closes #434

### DIFF
--- a/src/TLuaEngine.cpp
+++ b/src/TLuaEngine.cpp
@@ -82,6 +82,15 @@ void TLuaEngine::operator()() {
         }
     }
 
+    // now call onInitFinal after all states have completed their onInit
+    auto FinalFutures = TriggerEvent("onInitFinal", "");
+    WaitForAll(FinalFutures, std::chrono::seconds(5));
+    for (const auto& Future : FinalFutures) {
+        if (Future->Error && Future->ErrorMessage != BeamMPFnNotFoundError) {
+            beammp_lua_error("Calling \"onInitFinal\" on \"" + Future->StateId + "\" failed: " + Future->ErrorMessage);
+        }
+    }
+
     auto ResultCheckThread = std::thread([&] {
         RegisterThread("ResultCheckThread");
         while (!Application::IsShuttingDown()) {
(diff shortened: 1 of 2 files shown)


src/TLuaEngine.cpp
```diff
@@ -433,6 +442,7 @@ void TLuaEngine::EnsureStateExists(TLuaStateId StateId, const std::string& Name,
         auto DataPtr = std::make_unique<StateThreadData>(Name, StateId, *this);
         mLuaStates[StateId] = std::move(DataPtr);
         RegisterEvent("onInit", StateId, "onInit");
+        RegisterEvent("onInitFinal", StateId, "onInitFinal");
```
WiserTixx, line 445: Why was this line added?

WiserTixx (changes requested):

src/TLuaEngine.cpp
Kipstz: Well, to create the event?

WiserTixx: That is something up to the scripter so that the server doesn't implicitly call functions.

Kipstz: So why do it for “onInit”?

WiserTixx: I'm not entirely sure but it might be left over from the old scripting engine.

Kipstz: Well, I know I just followed what had already been done at the time. I think the implementation is fine. If in the future we have to redo the entire basis of the system, I would do it again, but I find it a bit strange to do things halfway. We might as well redo everything properly in the future.

WiserTixx: I think the best solution would be to remove both of those calls and see if anything breaks.

PR #453, opened 2025-12-24
Simplify FS.CreateDirectory return value

### LINKED ISSUE #187: [Feature Request] Only return error message from `FS.CreateDirectory()`
The Lua API function `FS.CreateDirectory()` currently returns a boolean that is false when an error occurred and the error message. It would be better to just return the error message or nil if there was no error. The boolean is just unnecessary and confusing.

## Example

```lua
local error_msg = FS.CreateDirectory("path")
if error_msg then
	print("Failed to create the directory:", error_msg)
end
```

---

PS: The documentation also wrongly states that the boolean is true when there was an error, whereas the inverse is actually true:

> Returns whether the operation had an error, and, if it did, an error message. This means that, if true is returned, an error occurred.

The example is also wrong:
```lua
local err, error_message = FS.CreateDirectory("data/mystuff/somefolder")

if err then
    print("failed to create directory: " .. error_message)
else
    -- do something with the directory
end
```

Kipstz:
Change FS.CreateDirectory to return nil on success instead of a pair. On error, returns the error message string directly.

Closes #187

### DIFF
--- a/include/LuaAPI.h
+++ b/include/LuaAPI.h
@@ -19,6 +19,7 @@
 #pragma once
 
 #include "TLuaEngine.h"
+#include <optional>
 #include <tuple>
 
 namespace LuaAPI {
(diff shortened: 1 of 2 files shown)


WiserTixx:
This is a nice change, however it will most likely break a lot of widely used server plugins. I'm not sure if it's worth the trade off.

Kipstz:
I see what you mean, but at some point we need to improve things. If it could cause problems, we should warn people for the future, but implement it anyway to make things cleaner.

WiserTixx:
I suspect warning people will not be sufficient in this case. Most people don't update their server plugins (often) so this change would be harming a lot of casual players for a slight correction in the lua api design.
Unless both versions of the function can be implemented without conflict this PR will have to wait for the lua api rework.

PR #455, opened 2025-12-24
Fix server crash when mods.json is empty

Kipstz:
Check file size before attempting to parse mods.json. 
Empty files are now gracefully ignored instead of causing a parse error.

### DIFF
--- a/src/TResourceManager.cpp
+++ b/src/TResourceManager.cpp
@@ -66,15 +66,20 @@ void TResourceManager::RefreshFiles() {
 
     nlohmann::json modsDB;
 
-    if (std::filesystem::exists(Path + "/mods.json")) {
-        try {
-            std::ifstream stream(Path + "/mods.json");
+    auto modsJsonPath = Path + "/mods.json";
+    if (std::filesystem::exists(modsJsonPath)) {
+        if (std::filesystem::file_size(modsJsonPath) > 0) {
+            try {
+                std::ifstream stream(modsJsonPath);
 
-            stream >> modsDB;
+                stream >> modsDB;
 
-            stream.close();
-        } catch (const std::exception& e) {
-            beammp_errorf("Failed to load mods.json: {}", e.what());
+                stream.close();
+            } catch (const std::exception& e) {
+                beammp_errorf("Failed to load mods.json: {}", e.what());
+            }
+        } else {
+            beammp_warn("Mods database file is empty, ignoring it");
         }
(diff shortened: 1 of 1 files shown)


src/TResourceManager.cpp
```diff
@@ -66,9 +66,10 @@ void TResourceManager::RefreshFiles() {
 
     nlohmann::json modsDB;
 
-    if (std::filesystem::exists(Path + "/mods.json")) {
+    auto modsJsonPath = Path + "/mods.json";
+    if (std::filesystem::exists(modsJsonPath) && std::filesystem::file_size(modsJsonPath) > 0) {
```
WiserTixx: Could a beammp_warn log be made when the file is empty? That way there's at least a trace of something not going as intended. Something along the lines of "Mods json database is empty, ignoring it"

src/TResourceManager.cpp
```diff
@@ -216,7 +217,7 @@ void TResourceManager::SetProtected(const std::string& ModName, bool Protected)
 
     auto modsDBPath = Application::Settings.getAsString(Settings::Key::General_ResourceFolder) + "/Client/mods.json";
 
-    if (std::filesystem::exists(modsDBPath)) {
+    if (std::filesystem::exists(modsDBPath) && std::filesystem::file_size(modsDBPath) > 0) {
```
WiserTixx: Could there be some kind of attempt to regenerate the file here? This way the protect command just gets completely disregarded with no info.

WiserTixx (changes requested):

WiserTixx:
Yes, this would technically solve the issue linked. However I'd like to figure out why that happened in the first place before closing it.

src/TResourceManager.cpp
```diff
@@ -220,20 +225,21 @@ void TResourceManager::SetProtected(const std::string& ModName, bool Protected)
...
 
-            std::fstream stream(modsDBPath);
-
-            stream >> modsDB;
+            if (std::filesystem::file_size(modsDBPath) > 0) {
+                std::ifstream inStream(modsDBPath);
+                inStream >> modsDB;
+                inStream.close();
+            } else {
+                beammp_warn("Mods database file is empty, regenerating it");
```
WiserTixx, line 233: I might be overlooking something here, but is the regeneration code missing?

WiserTixx (changes requested):

PR #459, opened 2026-01-11
Add Docker support and CI workflow for Docker image publish

enzofrnt:
This pull request introduces comprehensive Docker support for the project, enabling streamlined containerized builds, testing, and deployment. It adds a multi-stage `Dockerfile`, a Docker Compose configuration, a GitHub Actions workflow for automated multi-architecture image builds, and improves `.dockerignore` handling. Additionally, it refines the Git submodule logic in CMake for more robust and context-aware submodule management.

**Docker support and automation:**

* Added a multi-stage `Dockerfile` to build and package the server and its dependencies, including cache optimizations, debug symbol handling, and a non-root runtime user.
* Introduced `compose.yaml` for local development and deployment, mapping configuration, resources, and data directories, and exposing necessary ports.
* Added `.github/workflows/docker.yml` to automate Docker builds for both amd64 and arm64 architectures, supporting branch and tag triggers, and pushing images to GitHub Container Registry.
* Created a `.dockerignore` file to exclude unnecessary files and directories from Docker build context, improving build performance and reducing image size.

**Build system improvements:**

* Enhanced `cmake/Git.cmake` to update submodules only when in a real Git checkout, provide clearer status messages, and avoid errors when Git or `.git` is not present. (Which can append in Docker build)

---


enzofrnt:
Here is how it's look when build :
https://github.com/enzofrnt/BeamMP-Server/pkgs/container/beammp-server
https://github.com/enzofrnt/BeamMP-Server/actions/runs/20898778316


To try it out :

Create `compose.yaml` file :

```yml
services:
  beammp-server:
    image: ghcr.io/enzofrnt/beammp-server:docker-support
    container_name: beammp-server
    restart: unless-stopped
    ports:
      - "30814:30814/tcp"
      - "30814:30814/udp"
    volumes:
      # Mount configuration directory
      - ./config:/config
      # Mount resources directory (mods, plugins, etc.)
      - ./resources:/resources:ro
      # Mount working directory for logs and data
      - ./data:/app/data
    environment:
      - TZ=UTC
```

Start the container (it will automaticly download the Docker image):
`docker compose up -d`

OfficialLambdax:
Awesome PR! I would suggest one change and that is to install LuaRocks into the container as well. A handful of scripts require rock modules to run and many users have trouble setting it up which in return makes scripters trying to not require lua rocks. If a docker container had support for that at default then that would open up a whole new world

Dockerfile
```diff
@@ -0,0 +1,92 @@
...
+
+RUN useradd -m -u 1000 beammp && \
+    mkdir -p /app /app/data /config /resources && \
+    chown -R beammp:beammp /app /config /resources
+
+# /work/build-server is a BuildKit cache mount (not persisted in layers), so copy from /work/out.
+COPY --from=builder /work/out/BeamMP-Server /app/BeamMP-Server
+
+WORKDIR /app
+USER beammp
```
OfficialLambdax, line 96: Will this actually run if the system has no such user? Users created during build are temporary to my knowledge, so what happens once the container is run?

enzofrnt:
Let's do it <3 But I will not be able to try it.

Dockerfile
enzofrnt: We create/use a non-root user in a container to follow the principle of least privilege: if the application (or one of its dependencies) is compromised, the attacker won’t automatically get root inside the container, which reduces what they can change, read, or break (especially with mounted volumes).
Also, this is only about privileges inside the container (the container’s own filesystem/process namespace). It does not change the user on the host machine or grant any extra rights “outside” the container—those are controlled by the container runtime and host OS permissions.

OfficialLambdax: I see, that would mean the container is run as root but inside it it runs the entrypoint under the beammp user that only exists in the container. Did i get that right?

enzofrnt: You're right

enzofrnt:
Can I get some Lua examples I can try, along with a brief explanation of how they work?

Dockerfile
OfficialLambdax: Just a thought. What if the user of the docker where able to set the to be used user in the compose
```yaml
services:
  beammp-server:
    image: ghcr.io/enzofrnt/beammp-server:docker-support
    container_name: beammp-server
    restart: unless-stopped
    user: 1000:1000 # <<<<========
    ports:
      - "30814:30814/tcp"
      - "30814:30814/udp"
    volumes:
      # Mount configuration directory
      - ./config:/config
      # Mount resources directory (mods, plugins, etc.)
      - ./resources:/resources:ro
      # Mount working directory for logs and data
      - ./data:/app/data
    environment:
      - TZ=UTC
```

and then inside the dockerfile it adapts that user. That way the entire container is build and run under that user, no root implied anywhere

OfficialLambdax:
> Can I get some Lua examples I can try, along with a brief explanation of how they work?

Sure. Lets do an example with LuaSocket. You can install a rock like this `luarocks install luasocket`, just push it into a console.

Then create a folder inside the beammp servers Resources/Server folder. Name it anyway you want. In that new folder create a new file, name it `main.lua`. Inside it paste
```lua
local Socket = require("socket")
```
Reboot the server and done (:

if this throws an error the setup failed.

Dockerfile
enzofrnt: Yes, it’s possible to run Docker commands without being root if your account is in the docker group (that’s about who can control Docker on the host).
The `user:` line in `compose.yaml `only sets the `UID:GID` the container process runs as. It does not create or “adapt” a user inside the image; it just affects permissions (especially for bind-mounted volumes) and helps avoid files being written as root on the host.

OfficialLambdax:
> If you go into the container, install something, and then restart it, anything you just changed or installed will disappear unless it’s stored in a volume.

A simple container restart doesnt remove any data. Redeploying that container would do
```bash
docker compose down -- would remove!
docker compose stop -- just stops!
```

enzofrnt:
> > If you go into the container, install something, and then restart it, anything you just changed or installed will disappear unless it’s stored in a volume.
> 
> A simple container restart doesnt remove any data. Redeploying that container would do
> 
> ```shell
> docker compose down -- would remove!
> docker compose stop -- just stops!
> ```

yes, my bad ^^ I havedelete my comment.

But it’s a bit tricky to install things manually like that. Some documentation should provide instructions on how to properly create your own image, so you can update it and bring containers up/down without too many issues.

Starystars67:
Furthering your discussion there of luarocks inclusion, What if as part of the entrypoint script it checks an ENV value for the modules to install via luarocks?

This way it could install them at start and (maybe?) skip them if already installed without requiring the user to have to enter the container which is great for those hosting via the likes of a game panel or hosting provider.

toinopt:
> Furthering your discussion there of luarocks inclusion, What if as part of the entrypoint script it checks an ENV value for the modules to install via luarocks?
> 
> This way it could install them at start and (maybe?) skip them if already installed without requiring the user to have to enter the container which is great for those hosting via the likes of a game panel or hosting provider.

Along with this having a ENV for the LUA version to install might be useful too.

OfficialLambdax:
> But it’s a bit tricky to install things manually like that. Some documentation should provide instructions on how to properly create your own image, so you can update it and bring containers up/down without too many issues.

I believe the proper way to have a rock installed is by the script that requires it. eg (havent tested)
```lua
local is_ok, Socket = pcall(require, "socket")
if not is_ok then -- if module is not present then install it
   os.execute("luarocks install luasocket") -- will block until done
   Socket = require("socket")
end
```

Starystars67:
> > But it’s a bit tricky to install things manually like that. Some documentation should provide instructions on how to properly create your own image, so you can update it and bring containers up/down without too many issues.
> 
> I believe the proper way to have a rock installed is by the script that requires it. eg (havent tested)
> 
> ```lua
> local is_ok, Socket = pcall(require, "socket")
> if not is_ok then -- if module is not present then install it
>    os.execute("luarocks install luasocket") -- will block until done
>    Socket = require("socket")
> end
> ```

With this, are you thinking that this would be done from a script intended to be running from the BeamMP server or as a separate install script that is run before the BeamMP server is started?

OfficialLambdax:
> With this, are you thinking that this would be done from a script intended to be running from the BeamMP server or as a separate install script that is run before the BeamMP server is started?

By a BeamMP server script. That way a user wanting to use a public server script can just drop it into their server and it will set itself up. Otherwise the user would have to edit the compose again. It takes a setup out of the "How to install" instructions, right? As in its more convenient

enzofrnt:
I have a main.lua file, but nothing shows up in the server logs. Is this normal?

Could anyone help me figure out what’s going on on Discord?

OfficialLambdax:
To make a better example. Lets imagine we take this script https://github.com/OfficialLambdax/BeamMP-ChatHook/blob/main/Server/ChatHook/main.lua#L12

And lets imagine `local Log = require("libs/Log")` is a rock. (it in fact isnt, but lets imagine). Now the user just installed the script into their server and is running it for the first time. Now `require` will fail because it doesnt have `libs/Log` installed. And with the technic i posted above it could handle that error by it very self and install that dependency and rerequire it.

OfficialLambdax:
> I have a main.lua file, but nothing shows up in the server logs. Is this normal?
> 
> Could anyone help me figure out what’s going on on Discord?

Im out of time for today todo that. But lets try this just to ensure it worked
```lua
local Socket = require("socket")
print(Socket)
```

If this prints a list of functions it worked!

enzofrnt:
I’m not sure that what you’re asking for here is BeamMP’s responsibility for the moment.

Providing a clean Docker image with Lua installed, plus some documentation explaining how to install LuaRocks packages (and maybe how to build your own image for reuse), seems reasonable. But for now, BeamMP should focus on providing a working server.

Once the basic container is stable and used in production, we can talk about further improvements.

Dockerfile
OfficialLambdax: And that would be desirable, a user wouldnt have to be root to edit volume data for example. And the process wouldnt unnecessarily run as root either

enzofrnt: You don't need to be root to edit volume content.

OfficialLambdax:
> I’m not sure that what you’re asking for here is BeamMP’s responsibility for the moment.
> 
> Providing a clean Docker image with Lua installed, plus some documentation explaining how to install LuaRocks packages (and maybe how to build your own image for reuse), seems reasonable. But for now, BeamMP should focus on providing a working server.
> 
> Once the basic container is stable and used in production, we can talk about further improvements.

I differ from that thought. Any larger community that wants to advance on their project sooner or later will need rocks to enhance their servers. Be it to establish connections between containers or to the web or to add persistency for their data in terms of using any database. Having the official docker already include luarocks would make that alot easier for the users of it to make their desires work.

enzofrnt:
I agree about LuaRocks, but I disagree with adding any installation scripts or other kinds of setup tooling inside container for the moment.

OfficialLambdax:
> I agree about LuaRocks, but I disagree with adding any installation scripts or other kinds of setup tooling inside container for the moment.

Oh yeah i see the conversation got mixed up. There would be no installation scripts inside the container. The previous discussion was just about how a lua script could install its dependencies itself without the need of any installation script

enzofrnt:
It's seems to work. But not everything.
image
I'm not sure that the issue come from my work.

enzofrnt:
A simple Dockerfile like this should be sufficient to install dependencies and build your own custom image. That said, I’m not sure whether the issue I’m seeing above is caused by Docker.
Working on it here : https://github.com/enzofrnt/BeamMP-Server/tree/docker-suppor-lua-rocks

```Dockerfile
FROM ghcr.io/beammp/beammp-server:latest

USER root
RUN luarocks --lua-version=5.3 install luasocket
USER beammp
```

enzofrnt:
@OfficialLambdax The container can now easily support Lua scripting.
Could you share a simple script I can try out?

enzofrnt:
# Example of modification to add Lua dependencies:

```dockerfile
FROM ghcr.io/beammp/beammp-server:<tag>

USER root
RUN luarocks --lua-version=5.3 install luasocket
USER beammp
```

OfficialLambdax:
On my end the build of the image fails with
```
failed to solve: process "/bin/sh -c mkdir -p /work/out &&     cmake -S /work -B /work/build-server -G Ninja       -DCMAKE_TOOLCHAIN_FILE=/work/vcpkg/scripts/buildsystems/vcpkg.cmake       -DCMAKE_BUILD_TYPE=Release       -DCMAKE_EXE_LINKER_FLAGS=\"-Wl,--export-dynamic\"       -DCMAKE_CXX_FLAGS=\"-O3 -g -Wl,-z,norelro -Wl,--hash-style=gnu -Wl,-z,noseparate-code -ffunction-sections -fdata-sections -Wl,--gc-sections\"       -DBeamMP-Server_ENABLE_LTO=${ENABLE_LTO}     && cmake --build /work/build-server --parallel ${BUILD_PARALLEL} -t BeamMP-Server     && objcopy --only-keep-debug /work/build-server/BeamMP-Server /work/build-server/BeamMP-Server.debug     && strip --strip-unneeded /work/build-server/BeamMP-Server     && objcopy --add-gnu-debuglink=/work/build-server/BeamMP-Server.debug /work/build-server/BeamMP-Server     && install -m 0755 /work/build-server/BeamMP-Server /work/out/BeamMP-Server" did not complete successfully: exit code: 1
```

enzofrnt:
> On my end the build of the image fails with
> 
> ```
> failed to solve: process "/bin/sh -c mkdir -p /work/out &&     cmake -S /work -B /work/build-server -G Ninja       -DCMAKE_TOOLCHAIN_FILE=/work/vcpkg/scripts/buildsystems/vcpkg.cmake       -DCMAKE_BUILD_TYPE=Release       -DCMAKE_EXE_LINKER_FLAGS=\"-Wl,--export-dynamic\"       -DCMAKE_CXX_FLAGS=\"-O3 -g -Wl,-z,norelro -Wl,--hash-style=gnu -Wl,-z,noseparate-code -ffunction-sections -fdata-sections -Wl,--gc-sections\"       -DBeamMP-Server_ENABLE_LTO=${ENABLE_LTO}     && cmake --build /work/build-server --parallel ${BUILD_PARALLEL} -t BeamMP-Server     && objcopy --only-keep-debug /work/build-server/BeamMP-Server /work/build-server/BeamMP-Server.debug     && strip --strip-unneeded /work/build-server/BeamMP-Server     && objcopy --add-gnu-debuglink=/work/build-server/BeamMP-Server.debug /work/build-server/BeamMP-Server     && install -m 0755 /work/build-server/BeamMP-Server /work/out/BeamMP-Server" did not complete successfully: exit code: 1
> ```

How do you build it ?

OfficialLambdax:
> How do you build it ?

I took a copy of your branch and then simply `docker compose up`

enzofrnt:
> > How do you build it ?
> 
> I took a copy of your branch and then simply `docker compose up`

Have you init submodules ?

```bash
git clone --recursive <...>
```

If you already cloned:

```bash
git submodule update --init --recursive
```

If you prefer, the image is available here:
```bash
docker pull ghcr.io/enzofrnt/beammp-server:pr-2
```

.env.example
```diff
@@ -0,0 +1,12 @@
...
+BEAMMP_PORT=
+BEAMMP_MAX_CARS=
+BEAMMP_MAX_PLAYERS=
+BEAMMP_MAP=
+BEAMMP_NAME=
+BEAMMP_DESCRIPTION=
+BEAMMP_TAGS=
+BEAMMP_RESOURCE_FOLDER=
+BEAMMP_AUTH_KEY=
+BEAMMP_LOG_CHAT=
```
OfficialLambdax: It be great if this contained the default values
```
BEAMMP_DEBUG=false
BEAMMP_PRIVATE=true
BEAMMP_PORT=30814
BEAMMP_MAX_CARS=1
BEAMMP_MAX_PLAYERS=8
BEAMMP_MAP=/levels/gridmap_v2/info.json
BEAMMP_NAME=BeamMP Server
BEAMMP_DESCRIPTION="BeamMP Default Description"
BEAMMP_TAGS=
BEAMMP_RESOURCE_FOLDER=Resources
BEAMMP_AUTH_KEY=
BEAMMP_LOG_CHAT=true
```

compose.yaml
```diff
@@ -0,0 +1,21 @@
...
+  beammp-server:
+    build:
+      context: .
+      dockerfile: Dockerfile
+    image: ghcr.io/beammp/beammp-server
+    container_name: beammp-server
+    restart: unless-stopped
+    ports:
+      - "30814:30814/tcp"
+      - "30814:30814/udp"
```
OfficialLambdax: If the user set the port in the .env file it must be set here as well
```
    ports:
      - "${BEAMMP_PORT}:${BEAMMP_PORT}/tcp"
      - "${BEAMMP_PORT}:${BEAMMP_PORT}/udp"
```

compose.yaml
```diff
@@ -0,0 +1,21 @@
+services:
+  beammp-server:
+    build:
+      context: .
+      dockerfile: Dockerfile
+    image: ghcr.io/beammp/beammp-server
+    container_name: beammp-server
+    restart: unless-stopped
```
OfficialLambdax, line 8: consider adding the bridge network as well. There is rarely a need for a simple container to have its own network
```
network_mode: bridge
```

OfficialLambdax:
> If you prefer, the image is available here:
> 
> ```shell
> docker pull ghcr.io/enzofrnt/beammp-server:pr-2
> ```

Yes that helps. Im on a slow machine atm and just to get to the error took me a hour. This helps.
As expected the dirs the container creates are root owned and the server itself is even not able to write to them
```
beammp-server  | [12/01/26 11:45:15] [INFO] Custom config requested via commandline arguments: '/config/ServerConfig.toml'
beammp-server  | [12/01/26 11:45:15] [INFO] No config file found! Generating one...
beammp-server  | [12/01/26 11:45:15] [ERROR] Failed to create/write to config file: Permission denied
beammp-server  | [12/01/26 11:45:15] [ERROR] A fatal exception has occurred and the server is forcefully shutting down.
beammp-server  | [12/01/26 11:45:15] [ERROR] Failed to create/write to config file
```

A simple `chown` fixed it, but just for your info

.env.example
enzofrnt: Done ;)

compose.yaml
enzofrnt: Done ;)

compose.yaml
enzofrnt: I appreciate the suggestion, but I prefer to stick to the default Docker Compose behavior.

While it might seem like overkill for a single container, the default Compose network provides better isolation than the shared docker0 bridge. Also, ùusing network_mode: bridgeù disables the embedded DNS server, which would prevent service discovery if anyone ever decide to add a second container (like a database or a dashboard) to this stack later.

enzofrnt:
> > If you prefer, the image is available here:
> > ```shell
> > docker pull ghcr.io/enzofrnt/beammp-server:pr-2
> > ```
> 
> Yes that helps. Im on a slow machine atm and just to get to the error took me a hour. This helps. As expected the dirs the container creates are root owned and the server itself is even not able to write to them
> 
> ```
> beammp-server  | [12/01/26 11:45:15] [INFO] Custom config requested via commandline arguments: '/config/ServerConfig.toml'
> beammp-server  | [12/01/26 11:45:15] [INFO] No config file found! Generating one...
> beammp-server  | [12/01/26 11:45:15] [ERROR] Failed to create/write to config file: Permission denied
> beammp-server  | [12/01/26 11:45:15] [ERROR] A fatal exception has occurred and the server is forcefully shutting down.
> beammp-server  | [12/01/26 11:45:15] [ERROR] Failed to create/write to config file
> ```
> 
> A simple `chown` fixed it, but just for your info

Hmm, I’m not able to reproduce it. I don’t really understand why you’re running into this issue, are you running the container as root user on your local device?

OfficialLambdax:
> Hmm, I’m not able to reproduce it. I don’t really understand why you’re running into this issue, are you running the container as root user on your local device?

Your right that was my bad. I ran it with sudo. In the meantime ive played around with luarocks in this container. Where as a rock isnt installed in the image but via any script that requires it at runtime. This here is a working solution ive came up with

```lua

local function addLocalLuaRocksToPackagPath()
    local username = io.popen("whoami"):read()
    local path = string.format('/home/%s/.luarocks/share/lua/5.3/?.lua', username)
    local cpath = string.format('/home/%s/.luarocks/lib/lua/5.3/?.so', username)

    if not package.path:find(username .. '/.luarocks') then
        package.path = package.path .. ';' .. path
    end
    if not package.cpath:find(username .. '/.luarocks') then
        package.cpath = package.cpath .. ';' .. cpath
    end
end

local function requireRock(name, package_name)
    local is_ok, rock = pcall(require, name)
    if not is_ok then
        os.execute('luarocks install --local ' .. package_name)
        rock = require(name)
    end
    return rock
end

addLocalLuaRocksToPackagPath() -- called on state init

local Socket = requireRock("socket", "luasocket") -- try require, if fail then install
print(Socket)
```

enzofrnt:
> 
> In the meantime ive played around with luarocks in this container. Where as a rock isnt installed in the image but via any script that requires it at runtime. This here is a working solution ive came up with
> 
I'm not sure that's the best approach. I feel that installing packages on the fly isn't really a best practice; it seems much more robust to install them cleanly once during the build process.

That said, it's open for debate. We should see what the project maintainers think and which method they prefer to define as the 'standard' moving forward.

I do admit, however, that your method could effectively work within the current image. It might be a handy solution for 'Docker noobies' (beginners) who aren't comfortable building their own custom images.

By the way, do you understand how my current implementation works?

OfficialLambdax:
> I'm not sure that's the best approach. I feel that installing packages on the fly isn't really a best practice; it seems much more robust to install them cleanly once during the build process.

The idea is that a user of the image can just drag and drop a script they sourced from a third party into their server and it install the dependencies it needs itself. Without the creator having to explain that the user must now clone the repo, edit the dockerfile and build it themselfs. No it will just work out of the box. That makes it very convenient for beginners and advanced users alike.

If the image forced them to install rocks during the build process then anytime the user would want to use a new rock theyd have to update the entire image and recreate all containers using it. That be rather in the way then be productive.

> By the way, do you understand how my current implementation works?

Could you specify what implementation you mean?

enzofrnt:
That is actually how Docker is supposed to work. It's the standard practice to ensure stability. Beginners can indeed use your proposition for ease of use, but this doesn't mean we shouldn't provide a solution to do it 'the right way' for others. By 'my implementation', I meant the Dockerfile structure I proposed earlier by adding dependecies in custom Dockerfile to make your own image.

OfficialLambdax:
> That is actually how Docker is supposed to work. It's the standard practice to ensure stability. Beginners can indeed use your proposition for ease of use, but this doesn't mean we shouldn't provide a solution to do it 'the right way' for others. By 'my implementation', I meant the Dockerfile structure I proposed earlier by adding dependecies in custom Dockerfile to make your own image.

Yes its the standard to install everything during the build process, i just want to respect that most users dont know how to set things up and make it as convenient as possible for users to install server side mods. And its just very convenient to make on the fly changes (drag, drop, restart, done - compared to fully setting everything up again after one change).

You ment this?
```dockerfile
FROM ghcr.io/beammp/beammp-server:latest

USER root
RUN luarocks --lua-version=5.3 install luasocket
USER beammp
```

Yes thats totally fine!

enzofrnt:
Yes, we agree !

toinopt:
> That is actually how Docker is supposed to work. It's the standard practice to ensure stability. Beginners can indeed use your proposition for ease of use, but this doesn't mean we shouldn't provide a solution to do it 'the right way' for others. By 'my implementation', I meant the Dockerfile structure I proposed earlier by adding dependecies in custom Dockerfile to make your own image.

What about providing both options?
One is the gold image with just the basics without lua and another is the noobie version.

I say this a someone that that to learn how to build docker images by trial and error pretty much on my own, at one point I even gave up because I couldn't figure out all the LUA modules needed.

I understand how much easier it would be for a mod developer to just have a small instructions basically being, use this noobie image and add the files to the correct folder and it will install everything it needs on its own.

Building a new image is a non-trivial step that requires learning a lot of stuff and ideally have a CI-CD system.

enzofrnt:
@toinopt I said: “Beginners can indeed use your proposal for ease of use, but that doesn’t mean we shouldn’t provide a solution to do it ‘the right way’ for others.”

To clarify: the current image contains Luarock and Lua and can be modified by someone who has the necessary skills ("actually how Docker is supposed to work"). More beginner users can also use the solution proposed by Lambdax. Both will work in the current state of the pull request, and personally I don’t want to change it.

Only the maintainers of BeamMP-Server can say whether they want modifications or not.

enzofrnt:
Nothing new here ??

WiserTixx (changes requested):
Could all french text be translated to English and could the `BEAMMP_PROVIDER_DISABLE_CONFIG` env var be added set to true in the compose file?

enzofrnt:
@WiserTixx 
I've converted all French text to English.

Regarding `BEAMMP_PROVIDER_DISABLE_CONFIG`: I’m not sure why it needs to be in the compose file by default. If someone wants to rely only on environment variables for config, they can add this variable themselves. The current setup already provides a volume for the configuration file, so users can choose either a mounted config or env vars.

If the preferred approach for Docker is to use only environment variables for server config, I'd rather do it explicitly: remove the configuration volume from the compose file and set `BEAMMP_PROVIDER_DISABLE_CONFIG=true` in the Dockerfile so the image is env-only by default. That would make the intended use clear and avoid mixing "config file via volume" with "env-only" in the same example. (Note that someone who wants to use a config file can easily add a volume and override the `BEAMMP_PROVIDER_DISABLE_CONFIG` value.)

What do you prefer: adding the variable in the compose file as requested, or switching the Docker example to env-only (no config volume + variable in the Dockerfile)?

WiserTixx:
There's already an env file loaded in the compose and there's also an env example file. Environment variables take priority over the config so there is no point in having both. env's are imo much easier to work with compared to mounts or editing the container, so let's go with env only. However because the env vars are registered in the compose it makes more sense to have the disable config var also in there. That way the user can still switch to a mounted config without editing the image.

enzofrnt:
Ok I will do so.

enzofrnt:
Done : )

.env.example
```diff
@@ -0,0 +1,13 @@
...
+BEAMMP_MAX_CARS=1
+BEAMMP_MAX_PLAYERS=8
+BEAMMP_MAP=/levels/gridmap_v2/info.json
+BEAMMP_NAME=BeamMP Server
+BEAMMP_DESCRIPTION="BeamMP Default Description"
+BEAMMP_TAGS=
+BEAMMP_RESOURCE_FOLDER=Resources
+BEAMMP_AUTH_KEY=
+BEAMMP_LOG_CHAT=true
+BEAMMP_PROVIDER_DISABLE_CONFIG=true
```
WiserTixx: Could this env var be moved to the compose directly? It's not relevant here and if you want to use the config file you'd have to edit the compose anyway.

.env.example
```diff
@@ -0,0 +1,13 @@
+BEAMMP_DEBUG=false
```
WiserTixx, line 2: Could a comment be included here that mentions the official env guide?
https://docs.beammp.com/server/manual/#env

compose.yaml
```diff
@@ -0,0 +1,17 @@
...
+    ports:
+      - "${BEAMMP_PORT}:${BEAMMP_PORT}/tcp"
+      - "${BEAMMP_PORT}:${BEAMMP_PORT}/udp"
+    volumes:
+      # Mount resources directory (mods, plugins, etc.)
```
WiserTixx: Could this comment be removed?

Dockerfile
```diff
@@ -0,0 +1,100 @@
...
+    chown -R beammp:beammp /app /config /resources
+
+COPY --from=builder /work/out/BeamMP-Server /app/BeamMP-Server
+
+WORKDIR /app
+USER beammp
+
+EXPOSE 30814
+ENTRYPOINT ["/app/BeamMP-Server"]
+CMD ["--config=/config/ServerConfig.toml"]
```
WiserTixx: Could this command line argument be removed considering it's no longer required?

WiserTixx (changes requested):

enzofrnt:
Everythings done

enzofrnt:
no news ?

Starystars67:
Generally speaking looks good to me but I am no C++ dev so no clue if there are best practices missed or to follow.

Starystars67 (commented):
I cant comment for the Git.cmake and docker.yaml but the other files look fine to me.

PR #461, opened 2026-01-15
feature: BEAMMP_PROVIDER_DISABLE_MP_SET environment variable

TechStreetDev:
This pull request adds BEAMMP_PROVIDER_DISABLE_MP_SET variable (true/false/1/0) to disable MP::Set for providers.

---

### DIFF
--- a/include/Env.h
+++ b/include/Env.h
@@ -26,6 +26,7 @@ enum class Key {
     // provider settings
     PROVIDER_UPDATE_MESSAGE,
     PROVIDER_DISABLE_CONFIG,
+    PROVIDER_DISABLE_MP_SET,
     PROVIDER_PORT_ENV,
     PROVIDER_IP_ENV
 };
--- a/src/Env.cpp
+++ b/src/Env.cpp
@@ -36,6 +36,9 @@ std::string_view Env::ToString(Env::Key key) {
     case Key::PROVIDER_DISABLE_CONFIG:
         return "BEAMMP_PROVIDER_DISABLE_CONFIG";
         break;
+    case Key::PROVIDER_DISABLE_MP_SET:
+        return "BEAMMP_PROVIDER_DISABLE_MP_SET";
+        break;
     case Key::PROVIDER_PORT_ENV:
         return "BEAMMP_PROVIDER_PORT_ENV";
         break;
--- a/src/TLuaEngine.cpp
+++ b/src/TLuaEngine.cpp
@@ -22,6 +22,7 @@
 #include "CustomAssert.h"
 #include "Http.h"
 #include "LuaAPI.h"
+#include "Env.h"
 #include "Profiling.h"
 #include "TLuaPlugin.h"
 #include "sol/object.hpp"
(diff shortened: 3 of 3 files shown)


src/LuaAPI.cpp
```diff
@@ -307,6 +308,13 @@ std::pair<bool, std::string> LuaAPI::MP::RemoveVehicle(int PID, int VID) {
 }
 
 void LuaAPI::MP::Set(int ConfigID, sol::object NewValue) {
+    auto DisableMPSet = Env::Get(Env::Key::PROVIDER_DISABLE_MP_SET).value_or("false");
+    bool mDisableMPSet = DisableMPSet == "true" || DisableMPSet == "1";
```
WiserTixx: Could the result of this code be cached? A simple bool variable directly above the function in this will should do.

src/LuaAPI.cpp
```diff
@@ -307,6 +308,13 @@ std::pair<bool, std::string> LuaAPI::MP::RemoveVehicle(int PID, int VID) {
...
+    if (mDisableMPSet) {
+        beammp_info("This feature has been disabled by your server provider.");
```
WiserTixx: Could this be changed to a `beammp_lua_error` log and could the lua state name and details about the setting be added? e.g. `[ERROR] MyPlugin tried to call MP.Set to change setting 'Map' but this was blocked by your server provider.`

WiserTixx (changes requested):

src/LuaAPI.cpp
```diff
@@ -306,7 +307,35 @@ std::pair<bool, std::string> LuaAPI::MP::RemoveVehicle(int PID, int VID) {
...
+        case 5: return "Name";
+        case 6: return "Description";
+        case 7: return "InformationPacket";
+        default: return "Unknown";
+    }
+};
+
 void LuaAPI::MP::Set(int ConfigID, sol::object NewValue) {
+    if (mDisableMPSet) {
+        beammp_lua_error(
```
WiserTixx: Could the lua state name/plugin name be included instead of "A script"?
and could you use the **f**ormat version of the log functions to make the code cleaner? 
Example: <https://github.com/BeamMP/BeamMP-Server/blob/9fa99741597812be3925a284ad6d852ceb7ec141/src/LuaAPI.cpp#L157>

WiserTixx (changes requested):

src/LuaAPI.cpp
TechStreetDev: > Could the lua state name/plugin name be included instead of "A script"?

I could not figure out how to do this, and there is no other method which does this as far as I can tell (to copy from)

WiserTixx: I suppose you could move the check (and error message) to https://github.com/BeamMP/BeamMP-Server/blob/6f196aca647a9b34039a57602baaa9bbdd28f18a/src/TLuaEngine.cpp#L921 because the state name is available in that scope.
Something like:
```cpp
if (mDisableMPSet) {
   MP.set_function("Set", [this](int ConfigID, sol::object NewValue){
      beammp_lua_errorf("A script ({}) tried to call MP.Set to change setting '{}' but this was blocked by your server provider.", StateId, GetSettingName(ConfigID));
   });
} else {
   MPTable.set_function("Set", &LuaAPI::MP::Set);
}
```

TechStreetDev:
@WiserTixx done, commits also squashed.

WiserTixx (approved):

PR #466, opened 2026-01-20
Decrease latency

lionkor:
Disables Nagle's algorithm.

some reading:
- https://web.archive.org/web/20250917042737/https://docs.redhat.com/en/documentation/red_hat_enterprise_linux_for_real_time/10/html/optimizing_rhel_for_real_time_for_low_latency_operation/improving-network-latency-using-tcpnodelay
- https://en.wikipedia.org/w/index.php?title=Nagle%27s_algorithm&oldid=1310712676
- https://stackoverflow.com/a/19995579/14181711

---

### DIFF
--- a/src/TNetwork.cpp
+++ b/src/TNetwork.cpp
@@ -246,8 +246,8 @@ void TNetwork::TCPServerMain() {
                 continue;
             }
             // Disable Nagle's algorithm, equivalent to TCP_NODELAY.
-            // Nagle's algorithm trades beeter bandwidth efficiency for worse latency.
-            // Real time multi-layer games want lower latency.
+            // Nagle's algorithm trades better bandwidth efficiency for worse latency.
+            // Real time multi-player games want lower latency.
             ip::tcp::no_delay NoDelayOpt(true);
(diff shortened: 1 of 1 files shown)


src/TNetwork.cpp
```diff
@@ -245,6 +245,12 @@ void TNetwork::TCPServerMain() {
                 beammp_errorf("Failed to accept() new client: {}", ec.message());
                 continue;
             }
+            // Disable Nagle's algorithm, equivalent to TCP_NODELAY.
+            // Nagle's algorithm trades beeter bandwidth efficiency for worse latency.
+            // Real time multi-layer games want lower latency.
```
WiserTixx:
```suggestion
            // Nagle's algorithm trades better bandwidth efficiency for worse latency.
            // Real time multi-player games want lower latency.
```

WiserTixx (changes requested):

src/TNetwork.cpp
lionkor: Good catch

lionkor:
This also needs to happen in the Launcher otherwise it doesn't work as well (see references provided)

PR #474, opened 2026-02-23
RegisterEvent anonymous functions implementation (Additionally prevent CreateEventTimer() to register the same event more than one time)

### LINKED ISSUE #441: Lua Stack Corruption leads to x86 stack overflow
OS (windows, linux, ...): Linux, Debian 12
BeamMP-Server Version: v3.8.5, v3.9.0

**Describe the bug**
This bug occurred in CaRP.

I debugged the core dump, and essentially the bug is that the stack is in an unexpected state (either due to a threading-related data race, or because it's empty), and then some call across the C++-Lua boundary somewhere panics. This causes the LuaAPI::PanicHandler to be called, which in turn tries to access the stack. This panics Lua, calling the PanicHandler, and so on.

This eventually exhausts the (real) stack of the machine and crashes the server.

The same crash is reproducible with the reproduction steps below, but the CaRP codebase does NOT contain that code. The error is likely different, even though this reproduction might help fix the symptom.

I would carefully suggest running the server with some sanitizers like TSAN or valgrind in production.

**To Reproduce**
Steps to reproduce *similar* behavior:
1. Make a Lua plugin with the content `error(nil)`
2. Run the server

**Expected behavior**
A panic log and the server continues to run as usual, just without the errored plugin.

### LINKED ISSUE #449: Support anonymous functions and multiple handlers in MP.RegisterEvent
## Summary

  Currently, `MP.RegisterEvent` only accepts a function name as a string, which limits flexibility for plugin developers. This proposal adds support for:
  1. **Anonymous functions** - Pass functions directly instead of referencing by name
  2. **Multiple handlers** - Register multiple handlers for the same event

  ## Current Behavior

  ```lua
  -- Must define named function first
  function onPlayerJoin(pid)
      print("Player joined: " .. pid)
  end
  MP.RegisterEvent("onPlayerJoin", "onPlayerJoin")

  -- ❌ Cannot do this:
  MP.RegisterEvent("onPlayerJoin", function(pid)
      print("Player joined")
  end)

  -- ❌ Cannot register multiple handlers for same event

  Examples

  Backward Compatibility (existing syntax still works)

  function handleServerCommand(cmd, args)
      print("Executing command: " .. cmd)
  end
  MP.RegisterEvent("onConsoleInput", "handleServerCommand")

  New Anonymous Function Syntax

  MP.RegisterEvent("onChatMessage", function(playerID, playerName, message)
      if message:match("^/help") then
          MP.SendChatMessage(playerID, "Available commands: /help, /info")
      end
  end)

  Multiple Plugins, Same Event

  -- Plugin: AntiCheat
  MP.RegisterEvent("onPlayerJoin", function(playerID)
      checkPlayerBan(playerID)
  end)

  -- Plugin: WelcomeSystem
  MP.RegisterEvent("onPlayerJoin", function(playerID)
      sendWelcomeMessage(playerID)
  end)

  -- Plugin: Analytics
  MP.RegisterEvent("onPlayerJoin", function(playerID)
      logPlayerJoin(playerID)
  end)
  ```

  ✅ All three handlers execute when a player joins!

  Technical Approach

  1. Modify MP.RegisterEvent to accept sol::object (string OR function)
  2. If a function is passed, store it in _G with a generated unique name (e.g., __AnonymousHandler_onPlayerJoin_1)
  3. Use an atomic counter per state to ensure unique naming
  4. Register the generated name with the existing event system

  Benefits

  - Cleaner code: No need to define and name every handler function
  - Modularity: Multiple plugins can hook into the same event independently
  - Familiar pattern: Similar to how events work in Node.js, FiveM/CFX, and other frameworks
  - Backward compatible: Existing string-based registrations continue to work

  Additional Context

  I've implemented this as a Lua wrapper in my plugin (https://github.com/Kipstz/Tree-BeamMP-Plugin/blob/main/_tree/events.lua), but a native C++ implementation would be more performant and benefit all plugin developers.

  I'm willing to implement this feature and submit a PR if the approach is approved.

boubouleuh:
This PR implements the usage of the function directly inside RegisterEvent() while keeping the possibility of using the function name.
```lua
MP.RegisterEvent("Test",function()
Util.LogInfo("TEST")
end)
MP.TriggerLocalEvent("Test")
```
This needed A LOT of changes; since I'm still a beginner in C++, there are probably a couple of issues that I would be happy to fix (can't imagine that there aren't any).

If this gets merged, it would be better to merge it before #473 to prevent conflicts regarding the ``DontCallOnInit`` variable that isn't completely removed in this PR but is in #473.

ofc, I tested every changeand it seemed to work well (im just not sure about the ``queued`` lua console cmd)

Closes #449 
Closes #441 the panichandler is fixed in this PR
---

### DIFF
--- a/include/TLuaEngine.h
+++ b/include/TLuaEngine.h
@@ -203,7 +203,7 @@ public:
     [[nodiscard]] std::vector<std::shared_ptr<TLuaResult>> TriggerLocalEvent(const TLuaStateId& StateId, const std::string& EventName, ArgsT&&... Args) {
         std::unique_lock Lock(mLuaEventsMutex);
         beammp_event(EventName + " in '" + StateId + "'");
-        if (mLuaEvents.find(EventName) == mLuaEvents.end()) { // if no event handler is defined for 'EventName', return immediately
+        if (!mLuaEvents.contains(EventName)) { // if no event handler is defined for 'EventName', return immediately
             return {};
         }
         std::vector<std::shared_ptr<TLuaResult>> Results;
@@ -214,7 +214,7 @@ public:
         }
         return Results;
     }
-    std::vector<std::variant<sol::basic_protected_function<sol::basic_reference<true>>, std::string>> GetEventHandlersForState(const std::string& EventName, TLuaStateId StateId);
+    std::vector<std::variant<std::shared_ptr<sol::basic_protected_function<sol::basic_reference<true>>>, std::string>> GetEventHandlersForState(const std::string& EventName, TLuaStateId StateId);
     void CreateEventTimer(const std::string& EventName, TLuaStateId StateId, size_t IntervalMS, CallStrategy Strategy);
     void CancelEventTimers(const std::string& EventName, TLuaStateId StateId);
     sol::state_view GetStateForPlugin(const fs::path& PluginPath);
--- a/include/TLuaEngine.h
+++ b/include/TLuaEngine.h
@@ -36,6 +36,7 @@
 #include <set>
 #include <toml.hpp>
 #include <unordered_map>
+#include <utility>
 #include <vector>
 
 #define SOL_ALL_SAFETIES_ON 1
@@ -76,7 +77,7 @@ struct TLuaResult {
     std::string ErrorMessage;
     sol::object Result { sol::lua_nil };
     TLuaStateId StateId;
-    sol::object Function;
+    LuaFunction Function;
     std::shared_ptr<std::mutex> ReadyMutex {
         std::make_shared<std::mutex>()
     };
@@ -112,7 +113,7 @@ public:
     };
 
     struct QueuedFunction {
-        sol::object FunctionObject;
+        LuaFunction FunctionObject;
         std::shared_ptr<TLuaResult> Result;
         std::vector<TLuaValue> Args;
         std::string EventName; // optional, may be empty
@@ -167,9 +168,9 @@ public:
     void ReportErrors(const std::vector<std::shared_ptr<TLuaResult>>& Results);
     bool HasState(TLuaStateId StateId);
     [[nodiscard]] std::shared_ptr<TLuaResult> EnqueueScript(TLuaStateId StateID, const TLuaChunk& Script);
-    [[nodiscard]] std::shared_ptr<TLuaResult> EnqueueFunctionCall(TLuaStateId StateID, const sol::object& FunctionObject, const std::vector<TLuaValue>& Args, const std::string& EventName);
+    [[nodiscard]] std::shared_ptr<TLuaResult> EnqueueFunctionCall(TLuaStateId StateID, const LuaFunction& FunctionObject, const std::vector<TLuaValue>& Args, const std::string& EventName);
     void EnsureStateExists(TLuaStateId StateId, const std::string& Name, bool DontCallOnInit = false);
-    void RegisterEvent(const std::string& EventName, TLuaStateId StateId, const sol::object& FunctionObject);
+    void RegisterEvent(const std::string& EventName, TLuaStateId StateId, const LuaFunction& FunctionObject);
     /**
      *
      * @tparam ArgsT Template Arguments for the event (Metadata) todo: figure out what this means
@@ -182,7 +183,7 @@ public:
     [[nodiscard]] std::vector<std::shared_ptr<TLuaResult>> TriggerEvent(const std::string& EventName, TLuaStateId IgnoreId, ArgsT&&... Args) {
         std::unique_lock Lock(mLuaEventsMutex);
         beammp_event(EventName);
-        if (mLuaEvents.find(EventName) == mLuaEvents.end()) { // if no event handler is defined for 'EventName', return immediately
+        if (!mLuaEvents.contains(EventName)) { // if no event handler is defined for 'EventName', return immediately
             return {};
         }
 
@@ -213,7 +214,7 @@ public:
         }
         return Results;
     }
-    std::vector<sol::basic_object<sol::basic_reference<>>> GetEventHandlersForState(const std::string& EventName, TLuaStateId StateId);
+    std::vector<std::variant<sol::basic_protected_function<sol::basic_reference<true>>, std::string>> GetEventHandlersForState(const std::string& EventName, TLuaStateId StateId);
     void CreateEventTimer(const std::string& EventName, TLuaStateId StateId, size_t IntervalMS, CallStrategy Strategy);
     void CancelEventTimers(const std::string& EventName, TLuaStateId StateId);
     sol::state_view GetStateForPlugin(const fs::path& PluginPath);
@@ -226,7 +227,7 @@ public:
     std::vector<std::string> GetStateTableKeysForState(TLuaStateId StateId, std::vector<std::string> keys);
 
     // Debugging functions (slow)
-    std::unordered_map<std::string /*event name */, std::vector<sol::object> /* handlers */> Debug_GetEventsForState(TLuaStateId StateId);
+    std::unordered_map<std::string /*event name */, std::vector<LuaFunction> /* handlers */> Debug_GetEventsForState(TLuaStateId StateId);
     std::queue<std::pair<TLuaChunk, std::shared_ptr<TLuaResult>>> Debug_GetStateExecuteQueueForState(TLuaStateId StateId);
     std::deque<QueuedFunction> Debug_GetStateFunctionQueueForState(TLuaStateId StateId);
     std::vector<TLuaResult> Debug_GetResultsToCheckForState(TLuaStateId StateId);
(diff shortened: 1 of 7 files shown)


boubouleuh:
When trying to add more than one event timer it break tho idk how to fix 🫡

boubouleuh:
seems like i found the issues now it doesnt crash anymore

WiserTixx:
I've been testing this for a little bit, and it seems like registering more than 3 of the same events causes an error when you try to call them.
image

boubouleuh:
> I've been testing this for a little bit, and it seems like registering more than 3 of the same events causes an error when you try to call them. image

im looking into it, however interestingly it work if its triggered with an event timer

boubouleuh:
took me alot of time to debug everything that needed to be debugged but i think its good now, also i now use a shared_ptr for LuaFunction to prevent issues with the lua reference and it seems to work, and the events are properly cleared on hotreload now

src/TLuaEngine.cpp
```diff
@@ -309,28 +330,29 @@ void TLuaEngine::WaitForAll(std::vector<std::shared_ptr<TLuaResult>>& Results, c
         bool Cancelled = false;
         size_t ms = 0;
         std::set<std::string> WarnedResults;
+        std::string functionName = GetFunctionName(Result->Function);
```
SaltySnail: This should be pascal case (FunctionName)

src/TLuaEngine.cpp
```diff
@@ -1140,13 +1181,26 @@ void TLuaEngine::StateThreadData::operator()() {
...
+                auto& Args = TheQueuedFunction.Args;
+                auto& FnObject = TheQueuedFunction.FunctionObject;
                 Result->StateId = mStateId;
-                sol::state_view StateView(mState);
-                auto Fn = StateView[FnName];
+                lua_settop(mState, 0);
+                sol::main_protected_function Fn;
+                std::string profileName;
+                if (std::holds_alternative<std::string>(FnObject)) {
+                    auto name = std::get<std::string>(FnObject);
```
SaltySnail, line 1207: This should be pascal case (Name)

src/TLuaEngine.cpp
```diff
@@ -1140,13 +1181,26 @@ void TLuaEngine::StateThreadData::operator()() {
...
+                    Fn = StateView[name];
+                    profileName = name;
```
SaltySnail, line 1209: This should be pascal case (ProfileName)

src/TConsole.cpp
```diff
@@ -732,36 +732,37 @@ void TConsole::HandleLuaInternalCommand(const std::string& cmd) {
...
-        for (const auto& Function : FunctionsInOrder) {
-            if (Uniques.count(Function) == 0) {
-                Uniques.insert(Function);
-                if (FunctionsCount.at(Function) > 1) {
-                    Application::Console().WriteRaw("    " + Function + " (" + std::to_string(FunctionsCount.at(Function)) + "x)");
+        std::vector<LuaFunction> Uniques;
+        int Index = 0;
+        for (const LuaFunction Function : FunctionsInOrder) {
+            if (!(std::find(Uniques.begin(), Uniques.end(), Function) == Uniques.end())) {
+                std::string functionName = GetFunctionName(Function);
```
SaltySnail: This should be pascal case (FunctionName)

include/TLuaEngine.h
```diff
@@ -281,7 +283,7 @@ class TLuaEngine : public std::enable_shared_from_this<TLuaEngine>, IThreaded {
         std::thread mThread;
         std::queue<std::pair<TLuaChunk, std::shared_ptr<TLuaResult>>> mStateExecuteQueue;
         std::recursive_mutex mStateExecuteQueueMutex;
-        std::vector<QueuedFunction> mStateFunctionQueue;
+        std::deque<QueuedFunction> mStateFunctionQueue;
```
SaltySnail: Why change std::vector to std::deque?

SaltySnail (changes requested):
Nice work, small changes requested.

src/LuaAPI.cpp
```diff
@@ -436,7 +436,10 @@ void LuaAPI::MP::PrintRaw(sol::variadic_args Args) {
 }
 
 int LuaAPI::PanicHandler(lua_State* State) {
-    beammp_lua_error("PANIC: " + sol::stack::get<std::string>(State, 1));
+    const char* message = lua_tostring(State, -1);
```
lionkor: There is a way to do this with sol2, it'd be weird to mix raw Lua and sol2 IMO

src/TLuaEngine.cpp
```diff
@@ -428,23 +450,18 @@ void TLuaEngine::EnsureStateExists(TLuaStateId StateId, const std::string& Name,
         beammp_debug("Creating lua state for state id \"" + StateId + "\"");
         auto DataPtr = std::make_unique<StateThreadData>(Name, StateId, *this);
         mLuaStates[StateId] = std::move(DataPtr);
-        RegisterEvent("onInit", StateId, "onInit");
```
lionkor, line 431: Does it still call onInit even when it's not registered? This is behavior that people rely on so it'd be a major version increment

src/TLuaEngine.cpp
```diff
@@ -541,12 +563,20 @@ sol::table TLuaEngine::StateThreadData::Lua_TriggerLocalEvent(const std::string&
     // TODO: make asynchronous?
     sol::table Result = mStateView.create_table();
     int i = 1;
-    for (const auto& Handler : mEngine->GetEventHandlersForState(EventName, mStateId)) {
-        auto Fn = mStateView[Handler];
+    const auto Handlers = mEngine->GetEventHandlersForState(EventName, mStateId);
```
lionkor, line 566: Considering you changed the result type from a set to a vector, you can now have duplicates, right? I'm missing where that's considered here.

src/TLuaEngine.cpp
```diff
@@ -428,23 +450,18 @@ void TLuaEngine::EnsureStateExists(TLuaStateId StateId, const std::string& Name,
...
     std::unique_lock Lock(mLuaEventsMutex);
-    mLuaEvents[EventName][StateId].insert(FunctionName);
+    auto& Events = mLuaEvents[EventName][StateId];
+    if (const auto it = std::find(Events.begin(), Events.end(), FunctionObject); it == Events.end()) {
+        Events.push_back(std::move(FunctionObject));
+    }
 }
 
-std::set<std::string> TLuaEngine::GetEventHandlersForState(const std::string& EventName, TLuaStateId StateId) {
+std::vector<std::variant<std::shared_ptr<sol::basic_protected_function<sol::basic_reference<true>>>, std::string>> TLuaEngine::GetEventHandlersForState(const std::string& EventName, TLuaStateId StateId) {
```
lionkor: this now allows duplicate event handlers, as the invariant the set provides no longer holds. I assume this was done because it didn't compile otherwise; likely because the std::variant with these types does not implement std::hash, or something. That should be easy to fix by implementing std::hash<std::variant<...>> etc.? Or is it a deliberate choice to change this behavior? In the latter case it would need to be documented, probably, as it breaks assumptions, right? It's been a while so correct me if I'm wrong

src/TLuaEngine.cpp
```diff
@@ -831,8 +861,8 @@ TLuaEngine::StateThreadData::StateThreadData(const std::string& Name, TLuaStateI
     });
     MPTable.set_function("GetOSName", &LuaAPI::MP::GetOSName);
     MPTable.set_function("GetServerVersion", &LuaAPI::MP::GetServerVersion);
-    MPTable.set_function("RegisterEvent", [this](const std::string& EventName, const std::string& FunctionName) {
-        RegisterEvent(EventName, FunctionName);
+    MPTable.set_function("RegisterEvent", [this](const std::string& EventName, sol::main_protected_function FunctionObject) {
```
lionkor: I might have missed it, but doesn't this entirely remove the old way to call this? Why is the function type a std::variant, if it's no longer possible to pass it a string? Or am I missing another place where this function is re-added?

lionkor (commented):
I had a look since I was looking for the stability improvements, and saw a couple things. Hope it's okay :)

include/TLuaEngine.h
lionkor: I'm curious about this too. It doesn't seem to be used like a deque (double ended queue), right?

lionkor:
Also the vector to dequeue change should only make it faster if we pop from the front, if we pop from the back, we don't need to memcpy and it shouldn't make a difference which one is used.

src/TLuaEngine.cpp
boubouleuh: yes it still call it, this one was used for the luaconsole but it seem useless, its fully removed with the dontcalloninit things in another pr that i did tho i asked if it was really useless and it got approved so i assumed it really is

boubouleuh:
> Also the vector to dequeue change should only make it faster if we pop from the front, if we pop from the back, we don't need to memcpy and it shouldn't make a difference which one is used.

yes its better because its doing alot of ``mStateFunctionQueue.pop_front();`` i even compared the performance for both

src/TConsole.cpp
```diff
@@ -732,36 +732,37 @@ void TConsole::HandleLuaInternalCommand(const std::string& cmd) {
...
-        std::vector<std::string> FunctionsInOrder;
-        while (!QueuedFunctions.empty()) {
-            auto Tuple = QueuedFunctions.front();
-            QueuedFunctions.erase(QueuedFunctions.begin());
-            FunctionsInOrder.push_back(Tuple.FunctionName);
-            FunctionsCount[Tuple.FunctionName] += 1;
+        size_t FunctionsCount = QueuedFunctions.size();
+        std::vector<LuaFunction> FunctionsInOrder;
+        for (int i = 0; i < FunctionsCount; ++i) {
+            FunctionsInOrder.push_back(QueuedFunctions[i].FunctionObject);
```
lionkor, line 738: Is it possible the performance is better because you've removed the .erase()?

boubouleuh: tbh i dont know about the performance difference here, i tested the performance inside the StateFunctionQueue scope where i replaced the erase by a pop_front instead in TLuaEngine

PR #475, opened 2026-03-08
Harden malformed vehicle packet parsing

lynch-a:
## Summary
Reject malformed vehicle reset and paint packets before slicing structured payloads, and catch parser exceptions in the TCP client loop.

## Problem
Or:<pid>-<vid>:null and similarly malformed Op packets could reach substr(npos) in ParseVehicle, throwing std::out_of_range. On TCP, that exception could escape the client loop and kill the thread.

## Changes
- Guard Or reset packets when no { payload exists
- Guard Op paint packets when no [ payload exists
- Add a small helper and unit test for structured payload extraction
- Catch std::exception around GlobalParser in the TCP client loop and disconnect the offending client

## Verification
- git diff --check
- Full build/tests not run in this shell because cmake was not installed

### DIFF
--- a/src/TNetwork.cpp
+++ b/src/TNetwork.cpp
@@ -670,7 +670,13 @@ void TNetwork::TCPClient(const std::weak_ptr<TClient>& c) {
             Client->Disconnect("TCPRcv failed");
             break;
         }
-        mServer.GlobalParser(c, std::move(res), mPPSMonitor, *this, false);
+        try {
+            mServer.GlobalParser(c, std::move(res), mPPSMonitor, *this, false);
+        } catch (const std::exception& e) {
+            beammp_warnf("Failed to receive/parse packet via TCP from client {}: {}", Client->GetID(), e.what());
+            Client->Disconnect("Failed to parse packet");
+            break;
+        }
     }
 
     if (QueueSync.joinable())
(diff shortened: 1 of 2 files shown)


lynch-a:
All work implemented by Chat GPT 5.4 after I captured a PCAP of this nerd crashing my server 
image

lynch-a:
image

here's the packet FYI

pedrotski:
This is similar to what we pushed to all our clients earlier (excluding the parsing). I can confirm it fixed the crash for us on thousands of servers.

For example:

```
if (PID != -1 && VID != -1 && PID == c.GetID()) {
    auto BracePos = Data.find('{');
    if (BracePos == std::string::npos) {
        return;
    }
    Data = Data.substr(BracePos);
    LuaAPI::MP::Engine->ReportErrors(LuaAPI::MP::Engine->TriggerEvent("onVehicleReset", "", c.GetID(), VID, Data));
    Network.SendToAll(&c, StringToVector(Packet), false, true);
}
```

```
if (PID != -1 && VID != -1 && PID == c.GetID()) {
    auto BracketPos = Data.find('[');
    if (BracketPos == std::string::npos) {
        return;
    }
    Data = Data.substr(BracketPos);
    LuaAPI::MP::Engine->ReportErrors(LuaAPI::MP::Engine->TriggerEvent("onVehiclePaintChanged", "", c.GetID(), VID, Data));
    Network.SendToAll(&c, StringToVector(Packet), false, true);

    auto CarData = c.GetCarData(VID);
```

WiserTixx (approved):

PR #484, opened 2026-04-11
Fix client disconnection logic and add IP address retrieval and

### LINKED ISSUE #483: [BUG] Server crashes when connection is not closed properly
OS (windows, linux, ...): debian12
BeamMP-Server Version: v3.9.2

**Describe the bug**
When client crashes or `ALT` + `F4`, or any other case where the connection is not closed properly. The server might crash with a core dump.

**To Reproduce**
1. Go on a server, and close the launcher !

**Expected behavior**
The server does not crahes (I guess 😆)

**Logs**
Sorry, there is nothing logged, just a **SIGABRT**.

**Additional context**
The server crashes with a **SIGABRT** caused by an uncaught `boost::system::system_error` exception escaping the thread's boundary, triggering `std::terminate()`.

`src/TNetwork.cpp:652`:
```cpp
void TNetwork::DisconnectClient(TClient &c, const std::string &R)
{
    if (c.IsDisconnected()) return;
    // THIS LINE THROWS when the socket is already closed:
    std::string ClientIP = c.GetTCPSock().remote_endpoint().address().to_string();
    mClientMapMutex.lock();
    if (mClientMap[ClientIP] > 0) {
        mClientMap[ClientIP]--;
    }
    if (mClientMap[ClientIP] == 0) {
        mClientMap.erase(ClientIP);
    }
    mClientMapMutex.unlock();
    c.Disconnect(R);
}
```
**BUT**, the fix is not as simple as a `try / catch` because of the `mClientMap`, my best bet would be to save the client's ip inside of `TClient` at all times. If you have any other idea, please tell me.

Neptnium:
Fixes #483 that was created after the merge of #478.
I also took the freedom to switch the mutex locking to a lockguard for safety reasons.

---

### DIFF
--- a/src/Client.cpp
+++ b/src/Client.cpp
@@ -149,6 +149,11 @@ TClient::TClient(TServer& Server, ip::tcp::socket&& Socket)
     : mServer(Server)
     , mSocket(std::move(Socket))
     , mLastPingTime(std::chrono::high_resolution_clock::now()) {
+    boost::system::error_code ec;
+    auto ep = mSocket.remote_endpoint(ec);
+    if (!ec) {
+        mIP = ep.address().to_string();
+    }
 }
 
 TClient::~TClient() {
--- a/include/Client.h
+++ b/include/Client.h
@@ -79,6 +79,7 @@ public:
     [[nodiscard]] const ip::tcp::socket& GetTCPSock() const { return mSocket; }
     [[nodiscard]] std::string GetRoles() const { return mRole; }
     [[nodiscard]] std::string GetName() const { return mName; }
+    [[nodiscard]] const std::string& GetIP() const { return mIP; }
     void SetUnicycleID(int ID) { mUnicycleID = ID; }
     void SetID(int ID) { mID = ID; }
     [[nodiscard]] int GetOpenCarID() const;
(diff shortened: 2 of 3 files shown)


src/Client.cpp
```diff
@@ -149,6 +149,11 @@ TClient::TClient(TServer& Server, ip::tcp::socket&& Socket)
     : mServer(Server)
     , mSocket(std::move(Socket))
     , mLastPingTime(std::chrono::high_resolution_clock::now()) {
+    boost::system::error_code ec;
+    auto ep = mSocket.remote_endpoint(ec);
+    if (!ec) {
+        mIP = ep.address().to_string();
+    }
```
WiserTixx, line 156: Is there a reason why the error code isn't handled here?

WiserTixx (changes requested):

src/Client.cpp
Neptnium: 'ec' beeing the error code I don't see what you're saying, could you please explain ?

WiserTixx: If there is an error the ip is skipped and then the client construction is finished. However if the remote_endpoint call failed I don't see why we'd need to keep the client.  The error also isn't logged making things harder to debug.

Neptnium: Oh, well this is something I did not think about, do you want me to change it ?

SaltySnail:
I think this PR is deprecated by https://github.com/BeamMP/BeamMP-Server/pull/489
However we can see what the changes compare to when 489 is merged

src/Client.cpp
WiserTixx: Yes, that would be preferred.

PR #489, opened 2026-04-19
Stability Improvements

lionkor:
1. Issue: Broken QueueThread joining logic which relies on thread being `joinable` which is unrelated to doing `join`, leading to the case where the network thread crashes when it reaches the end of scope, due to a `std::terminate` in the dtor of `std::thread` (throws on destruction when unjoined). Fix: Using `std::jthread` which uses RAII to join on destruction.
2. Issue: Slow hardware or overloaded server can cause a client to disconnect and socket to close before `TNetwork::DisconnectClient` is called, but `IsDisconnected()` will still be false. In this case, `.remote_endpoint()` can fail, which throws an exception, which is not caught and thus `std::terminate`s the server. I observed this, even though it was only in extremely contrived scenarios, it still happened and crashed the server. Fix: Wrapping the whole block in a try/catch. Alternate fix would have been to pass an `error_code` but the result is the same. This fix is not quite enough though, a later fix resolves the remaining issue that this causes the `mClientMap` to ignore the disconnection. Also, all paths that do `if (c.IsDisconnected())` will fail to decrement the `mClientMap`, which isn't always the right behavior afaik. Fixed in another fix though.
3. Issue: Connection limiting ("DDoS protection") is broken as exceptions cause it to not decrement (and slowly fill up the `mClientMap`) in special cases. Mutexes are locked and unlocked manually which can (and will) lead to cases where the mutex is locked, an exception is thrown in the subsequent line (`address().to_string()` can throw, same with `.remote_endpoint()`, both of which are being called in the locked context without RAII unlocking). Fix: Replace manual map and mutex handling with a new class, `TConnectionLimiter`, and an associated "Guard" object `TConnectionLimiter::TGuard` which uses RAII to correctly keep track of IP-and-connection-count associations, the way the previous code was trying to do. This works across exceptions and other weird issues. Each connection's main thread now owns a guard, which, on destruction, decrements the counter. This way both the per-IP limits as well as the global limits are enforced. Also added some stats about this to the `status` command to ensure that server owners can observe this in action.
4. Issue: `.address().to_string()` can throw and is called even if `accept` failed, in `TNetwork::TCPServerMain`'s accept loop. I didn't observe this and it didn't cause crashes, but I was touching that part anyway. Fix: Explicitly handle the error case first, then get the IP, etc.
5. Issue: `ReadWithTimeout` spawned a new async context for each read, and then ran that context's event loop in a new thread. This means that, not only did every one of those reads SPAWN A THREAD(!!!), it also started an io context, which gets an fd, so this made DDoS arguably more effective, not less. 
6. Issue: Client disconnect can race due to being done on multiple threads (TOCTOU bug). For example `Looper` and a normal disconnect call can happen at the same time, because they check for `is_open` which can be true, and then change to false right after the check, causing a segfault in asio internals. Fix: Added an atomic compare-and-swap (CAS) mechanic that acts like a lock for the socket disconnect/close, and adjusted other places that checked `is_open`.
8. Issue: Lua panic calls the panic handler, and if the error supplied in the panic is not a string, or is otherwise invalid, it will trigger another panic within the panic handler. This continues and eventually crashes the program in one of many fun ways. Fix: Use raw lua functions to check if the top of the stack is a string, and only then print it, otherwise print that there was a panic and leave it at that.
9. Issue: `error()` crashes the server, due to `sol::error`'s constructor expecting a `std::string` (`lua_tostring` or `__tostring` meta method), which doesn't exist if the error is, for example, `nil`. I reported this to sol2, but it might be an issue only in this older version we're using. Fix: Fixed as part of the next issue:
10. Issue: `TLuaResult` was used/accessed from multiple threads, including `sol::object` accessed from multiple threads. This lead to each access of a `TLuaResult::Result` accessing the Lua stack of that state (from outside that state's thread, which is unsafe). This consistently lead to issues and sometimes crashes. Fix: `TLuaResult` now always marshals results into a detached result variant. This allocates, but this is unlikely to impact the hot paths, as most results will be empty or have primitive types. **UPDATE 2026-04-29:** This fix caused some other issues, so I added a result type that has no result value, only a status. That seems to help.
11. Issue: HTTP retains a curl handle per thread and never cleans them up. With one new thread per client, each doing an auth request at least, this quickly exhausts all file descriptors. This manifests itself as **dns resolutions failing**, as the server fails to open a socket to send a DNS query. Fix: The HTTP code now retains a pool of reusable handles, which clean up automatically via RAII. I tried to build this in a way that doesn't modify the code too much, so I kept it global and static.
12. Issue: Crash when accessing an expired `std::weak_ptr<TClient>`. This can happen when we check for `.expired()`, and then `.lock()`, which is, of course, another TOCTOU (time of check vs time of use) bug. What youre SUPPOSED to do instead, is locking, which always returns a `std::shared_ptr<>`, and then check `std::shared_ptr<>::operator bool`. An expired `std::weak_ptr` will return a default-constructed `std::shared_ptr`, which evaluates to `false` when converted to `bool`. Fix: Replaced all uses of `.expired()` and other such checks with the correct pattern. This was a lot of search and manual replace :D.

I used LLMs to help with writing unit-tests, but those do not compile into the final executable anyway. If this is undesired, I'm happy to remove that code.

---

### DIFF
--- a/src/LuaAPI.cpp
+++ b/src/LuaAPI.cpp
@@ -440,7 +440,16 @@ void LuaAPI::MP::PrintRaw(sol::variadic_args Args) {
 }
 
 int LuaAPI::PanicHandler(lua_State* State) {
-    beammp_lua_error("PANIC: " + sol::stack::get<std::string>(State, 1));
+    // panic path: use raw lua c api only; sol2 conversions can raise lua_error
+    // and re-enter panic recursively.
+    const int ErrorType = lua_type(State, -1);
+    if (ErrorType == LUA_TSTRING) {
+        const char* Message = lua_tostring(State, -1);
+        beammp_lua_error(std::string("PANIC: ") + (Message ? Message : "(null)"));
+    } else {
+        const char* ErrorTypeName = lua_typename(State, ErrorType);
+        beammp_lua_error(std::string("PANIC: non-string error object (") + (ErrorTypeName ? ErrorTypeName : "unknown") + ")");
+    }
     return 0;
 }
 
--- a/src/LuaAPI.cpp
+++ b/src/LuaAPI.cpp
@@ -143,22 +143,23 @@ static inline std::pair<bool, std::string> InternalTriggerClientEvent(int Player
         return { true, "" };
     } else {
         auto MaybeClient = GetClient(LuaAPI::MP::Engine->Server(), PlayerID);
-        if (!MaybeClient || MaybeClient.value().expired()) {
-            beammp_lua_errorf("TriggerClientEvent invalid Player ID '{}'", PlayerID);
-            return { false, "Invalid Player ID" };
-        }
-        auto c = MaybeClient.value().lock();
+        if (MaybeClient) {
+            if (auto c = MaybeClient.value().lock()) {
+                if (!c->IsSyncing() && !c->IsSynced()) {
+                    return { false, "Player hasn't joined yet" };
+                }
 
-        if (!c->IsSyncing() && !c->IsSynced()) {
-            return { false, "Player hasn't joined yet" };
+                if (!LuaAPI::MP::Engine->Network().Respond(*c, StringToVector(Packet), true)) {
+                    beammp_lua_errorf("Respond failed, dropping client {}", PlayerID);
+                    LuaAPI::MP::Engine->Network().ClientKick(*c, "Disconnected after failing to receive packets");
+                    return { false, "Respond failed, dropping client" };
+                }
+                return { true, "" };
+            }
         }
 
-        if (!LuaAPI::MP::Engine->Network().Respond(*c, StringToVector(Packet), true)) {
-            beammp_lua_errorf("Respond failed, dropping client {}", PlayerID);
-            LuaAPI::MP::Engine->Network().ClientKick(*c, "Disconnected after failing to receive packets");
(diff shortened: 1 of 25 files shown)


src/LuaAPI.cpp
```diff
@@ -189,21 +192,26 @@ std::pair<bool, std::string> LuaAPI::MP::SendChatMessage(int ID, const std::stri
...
+                    Result.first = false;
+                    Result.second = "Player still syncing data";
+                    return Result;
+                }
+                if (LogChat) {
+                    LogChatMessage("<Server> (to \"" + c->GetName() + "\")", -1, Message);
+                }
+                if (!Engine->Network().Respond(*c, StringToVector(Packet), true)) {
+                    beammp_errorf("Failed to send chat message back to sender (id {}) - did the sender disconnect?", ID);
+                    // TODO: should we return an error here?
```
SaltySnail, line 207: Returning an error here makes sense in my opinion as that is how it is also handled when the player is still syncing

SaltySnail: Although I see now that the comment has moved instead of being introduced, so I won't hold up the PR for this

SaltySnail (approved):
LGTM, well done I like the RAII connection limiter

PR #490, opened 2026-04-26
Add asynchronous HTTP handling with HttpAsync module

cocorico8:
#### Description
This PR introduces a comprehensive, non-blocking asynchronous networking suite to the server-side Lua API, consisting of the `AsyncHttp` and `AsyncWebSocket` modules. It provides modders with high-performance tools for external integrations while strictly protecting the server's main thread and system resources.

#### Key Features:
*   **Full REST & WebSocket Support:** Implements `GET`, `POST`, `PUT`, `PATCH`, `DELETE`, `HEAD`, and full-duplex **WebSockets** (`ws://` and `wss://`).
*   **Request Handles & Cancellation:** All requests return a handle allowing for manual cancellation via `:Cancel()`, which immediately severs the underlying TCP socket to reclaim resources.
*   **Smart Payload Handling:** Automatically detects Lua tables and encodes them to JSON, defaulting to `application/json` where appropriate.
*   **Advanced Networking:** 
    *   **Separate Timeouts:** Granular control over Connection and Read timeouts to prevent thread-hangs on dead endpoints.
    *   **Standards Compliant Headers:** Support for multi-value headers (e.g., multiple `Set-Cookie` fields) and automatic lowercase header normalization for easy Lua access.
*   **File I/O:** Streaming `Download` to disk and `PostFile` for multipart form-data uploads.
*   **Universal Progress Tracking:** Real-time progress callbacks supported for all HTTP methods and downloads.

#### Stability & Safety Implementation:
*   **Dynamic Resource Scaling:** The module automatically scales the internal `httplib::ThreadPool` and WebSocket quotas based on the host's hardware (CPU thread count).
*   **Plugin Sandboxing:** Implements per-plugin rate limiting (quotas) for both HTTP requests and active WebSocket threads. This prevents a single buggy or malicious script from exhausting the server's network capacity.
*   **Memory Management:** 
    *   Utilizes a thread-safe result queue processed during the server's `Update` cycle to ensure Lua callbacks only execute on the main thread.
    *   Robust `CleanupState` and `Abandon` logic ensures all Lua Registry references are freed and background threads are terminated immediately when a script reloads or stops.
*   **Thread Safety:** Comprehensive use of mutexes and atomic booleans across the bridge to prevent race conditions during async operations.

---

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -31,6 +31,7 @@ set(PRJ_HEADERS
     include/Defer.h
     include/Environment.h
     include/Http.h
+    include/HttpAsync.h
     include/IThreaded.h
     include/Json.h
     include/LuaAPI.h
@@ -60,6 +61,7 @@ set(PRJ_SOURCES
     src/Common.cpp
     src/Compat.cpp
     src/Http.cpp
+    src/HttpAsync.cpp
     src/LuaAPI.cpp
     src/SignalHandling.cpp
     src/TConfig.cpp
--- a/include/HttpAsync.h
+++ b/include/HttpAsync.h
@@ -32,7 +32,7 @@ namespace HttpAsync {
         
         int status;
         std::string body;
-        std::map<std::string, std::string> headers;
+        std::map<std::string, std::vector<std::string>> headers; 
         
         long long current;
         long long total;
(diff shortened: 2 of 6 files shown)


cocorico8:
currently there's a bug in the version of cpp-httplib, I would need to update the baseline to get rid of it to avoid using hacky solution, should i make another pr to update the vcpkg baseline (tried in a debian 12 vm that it still compiles and it does)

PR #495, opened 2026-06-18
Add "RegisterIPv6" setting

iguanajuice:
these 2 lines of code are seemingly useless; removing them makes my ipv6 only beammp container work

---


WiserTixx:
Those lines are not useless, not everything supports ipv6 (in beammp and for users). Without this line all servers on the server list that support ipv6 will change their listing ip to ipv6. I do not mind this being an option to server operators but I'd prefer it to be toggled via an environment variable. E.g. `BEAMMP_HEARTBEAT_IPV6`, along with a warning in the console that this might make the server unavailable to some users joining via the public list.

iguanajuice:
libcurl is smart enough to use both ipv4 and ipv6. i've just now tested beammp server in an ipv4 only container, and it has no problem connecting to the backend with this patch.

so yes, im pretty sure restricting to ipv4 only is useless.

WiserTixx:
That's not my point. I never said curl was incapable of handling that. 
Ipv6 doesn't work for all clients and the backend only lists (and knows) the IP which the server heartbeats from. With this change that would always be ipv6 if available on the host). If a server owners who uses a machine that supports ipv6 were to upgrade to a version with this change it would render their server unusable for most/a large proportion of the player base, without ever changing anything on their side.
I do not want to force server owners to disable ipv6 on their host or spin up an ipv4 only container for every server that works just fine at the moment. Especially considering the fact that a lot of end users simply do not have internet service which supports ipv6.

iguanajuice:
I'm failing to replicate the scenario you are describing. I created a dual-stack container while also dropping all out-going IPv6 packets (the container/server has an ipv6 address, but one that doesn't work). The server *still* works. 

So unless I'm yet again misinterpreting you, it seems that the server has no problem falling back to ipv4 when ipv6 is assigned but not working.

WiserTixx:
I'm not talking about the server, it works just fine. When a server registers on the backend with ipv6, and a client without ipv6 support, either because the BeamMP mod/Launcher doesn't support it, or because the user's router/ISP doesn't support it. It will fail, without prior notice. For this to be resolved sizeable changes would need to be made to the backend.
It prevent servers from becoming inaccessible for part of the playerbase this change must be opt-in, and with warning. Just removing that line will make all servers that support ipv6 use it by default for backend requests, thus breaking their public server list entry for some/a lot of players.

iguanajuice:
I think I understand now.

The best solution would be to attempt to register both address families with the backend, so that a dual-stack server will work with as many users as possible without further configuration.

I would take a crack at this myself, but I don't think the backend is open-source.

iguanajuice:
But for now, I've added a new setting to solve my problem. Default behaviour is identical to pre-patch.

PR #498, opened 2026-06-23
Add region tld

SaltySnail:
Adds region support for the ServerConfig.toml, which controls the top level domain being used

---

### DIFF
--- a/src/Common.cpp
+++ b/src/Common.cpp
@@ -35,6 +35,7 @@
 #include "Compat.h"
 #include "CustomAssert.h"
 #include "Http.h"
+#include "RegionHandler.h"
 
 void Application::RegisterShutdownHandler(const TShutdownHandler& Handler) {
     std::unique_lock Lock(mShutdownHandlersMutex);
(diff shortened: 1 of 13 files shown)


src/Common.cpp
```diff
@@ -73,6 +73,38 @@ std::string Application::ServerVersionString() {
...
+}
+
+std::string Application::GetBackendUrlForSocketIO()
+{
+    return "https://backend." + RegionToTopLevelDomain(Settings.getAsString(Settings::Key::General_Region));
+}
+
+std::string Application::RegionToTopLevelDomain(const std::string region)
+{
+    if (region == "Restricted") {
```
WiserTixx: https://github.com/BeamMP/BeamMP-Launcher/pull/256#discussion_r3470422570

src/Settings.cpp
```diff
@@ -36,6 +36,7 @@ Settings::Settings() {
         { General_Debug, false },
         { General_AllowGuests, true },
         { General_InformationPacket, true },
+        { General_Region, std::string("Global")},
```
WiserTixx: Couldn't this just be an empty string?

src/TConfig.cpp
```diff
@@ -52,6 +52,8 @@ static constexpr std::string_view StrTags = "Tags";
 static constexpr std::string_view EnvStrTags = "BEAMMP_TAGS";
 static constexpr std::string_view StrResourceFolder = "ResourceFolder";
 static constexpr std::string_view EnvStrResourceFolder = "BEAMMP_RESOURCE_FOLDER";
+static constexpr std::string_view StrRegion = "Region";
```
WiserTixx: It might be better to name this something along the lines of `BackendRegion` or `BeamMPBackendRegion` so that users don't confuse it with their server region or something along those lines.

src/TConfig.cpp
```diff
@@ -151,6 +153,7 @@ void TConfig::FlushToFile() {
     data["General"][StrMap.data()] = Application::Settings.getAsString(Settings::Key::General_Map);
     data["General"][StrDescription.data()] = Application::Settings.getAsString(Settings::Key::General_Description);
     data["General"][StrResourceFolder.data()] = Application::Settings.getAsString(Settings::Key::General_ResourceFolder);
+    data["General"][StrRegion.data()] = Application::Settings.getAsString(Settings::Key::General_Region);
```
WiserTixx: A comment in the config file here explaining this setting would be beneficial to avoid confusion, though I'm not sure if having it in the config file by default is very efficient (users might change this on their own without knowing what it is and then support has to deal with that_. 
It might be better suited as a more "hidden" setting that you can add as a line to the config, and only then will it also be written out here.

WiserTixx (approved):

src/Settings.cpp
SaltySnail: Fixed in new commits

src/TConfig.cpp
SaltySnail: Fixed in new commits

src/TConfig.cpp
SaltySnail: Fixed in new commits

WiserTixx (changes requested):
Same comments as on the launcher side

WiserTixx:
Whenever I start the server initially I get a `[WARN] Value 'General.BackendRegion' has unexpected type, expected type 'string'`

PR #499, opened 2026-07-05
Fix silent truncation of UDP datagrams over 1024 bytes

IrPgFKS0:
### Problem

`TNetwork::UDPRcvFromClient` receives into a fixed `std::array<char, 1024>`. UDP `recvfrom()` **truncates** any datagram larger than the buffer — the tail is silently lost. A compressed (`ABG:`) position packet from a client with several moving vehicles can exceed 1 KB, so those updates were silently dropped (the truncated payload fails to decompress/parse), and other players saw those vehicles stutter or warp.

### Fix

Size the receive buffer to the UDP maximum (65535). One-line change plus a comment; the function already returns only the received byte count, so there is no behavior change for small datagrams.

### How this was found

Diagnosed while profiling remote-vehicle drift on a LAN fork of BeamMP (position updates from multi-vehicle clients disappearing under load); the fix has been running in that fork for weeks of sessions.

### Transparency

This fix comes from an AI-assisted fork: the bug was found and the patch written with the help of an AI coding tool (Claude), then tested by a human in real multiplayer sessions. Given this project's policy on AI-generated code, it's submitted as a **draft** for the maintainers to decide — happy to close it if that's not wanted, or for a maintainer to re-implement the one-liner independently.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/src/TNetwork.cpp
+++ b/src/TNetwork.cpp
@@ -1356,7 +1356,10 @@ bool TNetwork::UDPSend(TClient& Client, std::vector<uint8_t> Data) {
 }
 
 std::vector<uint8_t> TNetwork::UDPRcvFromClient(boost::asio::ip::udp::endpoint& ClientEndpoint) {
-    std::array<char, 1024> Ret { };
+    // Sized to the UDP maximum: recvfrom() truncates a datagram that exceeds the buffer
+    // (the tail is silently lost). A compressed position packet from a client with several
+    // moving vehicles can exceed 1024 bytes, so such updates were silently dropped.
+    std::array<char, 65535> Ret { };
     boost::system::error_code ec;
     const auto Rcv = mUDPSock.receive_from(boost::asio::mutable_buffer(Ret.data(), Ret.size()), ClientEndpoint, 0, ec);
     if (ec) {


WiserTixx:
Is there any evidence that this improves position accuracy? Are there numbers or is this all just anecdotal?

IrPgFKS0:
I think what you are asking is if there is evidence of datagrams being dropping and if those dropped datagrams caused a noticeable measurable impact?  If so this is partly anecdotal, but more importantly was only noticed and tested on a LAN network so I would assume the conditions where this could occur and would be noticeable over the internet would be negligible since there would be many other reasons for lost UDP datagrams that would outweigh what this PR is fixing.   The reason I submitted this is just to ensure the mods own receiver isn't one of those reasons, but I can gather more evidence if needed.

WiserTixx:
Have you seen udp datagrams over 1 KB there were sent to the server? If so, how many and which packet codes/types were seen and subsequently cut off? On a standard BeamMP setup. (Excluding debug code to log this occurrence)

IrPgFKS0:
I will test specifically on vanilla BeamMP and see if I can reproduce over an internet session and report back.

IrPgFKS0:
Closing this — I went back to verify the premise before pushing it any further, and it doesn't hold up.

`ServerSend` in the launcher already forces anything with `compressBound(len) > 1024` onto TCP. That caps a UDP payload at 1011 bytes and the resulting datagram at 1030, and only if the payload were incompressible — position JSON isn't. That gate has been in place since f2b3454 (2024-06-22), so it was already there when I filed this. `ServerSend` also `abort()`s on a `Zp` packet over 500 bytes, which is upstream's own assertion that these stay small, and the measured average packet in a 19-vehicle session on my fork was ~314 bytes.

I also have no log evidence of the symptom I described: zero `UDP recvfrom() failed` lines across months of server logs. That absence is meaningful on Windows specifically, where an oversized datagram fails loudly with `WSAEMSGSIZE` rather than truncating silently.

The one loophole I could think of doesn't help either: clients predating the gate had no size limit, but the server kicks anything below `ClientMinimumVersion` 2.7.0 (released 2025-10-20), sixteen months after the gate shipped — so every client that can connect has it.

So the "stutter or warp" I attributed to this isn't attributable to it, and the one-liner isn't worth your review time on its own merits. Apologies for the noise.

For what it's worth, the change still has a narrow defensive value that I'd frame completely differently: a malformed or non-stock client *can* send a datagram larger than the buffer, and on Linux that truncates silently and hands the parser a corrupt packet rather than failing loudly. If that's ever interesting it belongs bundled with other input-validation guards, not as a bug fix — I'm not opening that as a PR unless you'd want it.

BeamMP-Server#500 is unaffected by any of this and still stands on its own.

PR #501, opened 2026-07-11
Fix server hang/crash on shutdown (#491)

### LINKED ISSUE #491: Issue on Linux, needs to kill server even after "EXIT" for it to properly stop
OS (windows, linux, ...): Linux (was only tested on debian)
BeamMP-Server Version: Latest (v3.9.3)

**Describe the bug**
In Pteoradctyl, when a server is stopped, it is doing the stopping process, but it never ends, you need to KILL it or send a console input. (See vids below)

**To Reproduce**
https://github.com/user-attachments/assets/30b7a166-cd49-4263-a04b-bc65e76a25a7
https://github.com/user-attachments/assets/a70c5dc5-bd29-4ac2-b1d8-2cecd225d907

**Expected behavior**
The server to stop correctly, and no need to kill it.

vladmaksimenko-0:
I run my server on Pelican and kept having to force-kill it on shutdown (same as [#491](https://github.com/BeamMP/BeamMP-Server/issues/491)). It finishes shutting down all its subsystems and prints Shutdown., then just hangs. One time instead of hanging it crashed with this:

```
terminate called after throwing an instance of 'std::logic_error'
  what():  Undefined key accessed in Settings::getAsBool
```

There's a commented-out std::exit at the end of GracefullyShutdown() that looked like it was meant to handle this, so I tried exiting there with std::_Exit(0) after the shutdown handlers run. That makes it shut down instantly and cleanly for me now.

I'm not a C++ dev so I might be missing the deeper cause. Traced it with some help from AI, but I reproduced and tested the fix on my own server.

Fixes [#491].

### DIFF
--- a/src/Common.cpp
+++ b/src/Common.cpp
@@ -66,7 +66,7 @@ void Application::GracefullyShutdown() {
         beammp_info("Subsystem " + std::to_string(i + 1) + "/" + std::to_string(mShutdownHandlers.size()) + " shutting down");
         mShutdownHandlers[i]();
     }
-    // std::exit(-1);
+    std::_Exit(0);
 }
 
 std::string Application::ServerVersionString() {


src/Common.cpp
```diff
@@ -66,7 +66,7 @@ void Application::GracefullyShutdown() {
         beammp_info("Subsystem " + std::to_string(i + 1) + "/" + std::to_string(mShutdownHandlers.size()) + " shutting down");
         mShutdownHandlers[i]();
     }
-    // std::exit(-1);
+    std::_Exit(0);
```
WiserTixx, line 69: Why `std::_Exit` instead of `std::exit`?

WiserTixx (changes requested):

PR #503, opened 2026-08-04
Add time ping handling

WiserTixx:
client -> server `t[8 bytes]`
server -> client `t[client's 8 bytes][server time 8 bytes]`

---

### DIFF
--- a/src/TServer.cpp
+++ b/src/TServer.cpp
@@ -123,6 +123,12 @@ TEST_CASE("GetPidVid") {
 TServer::TServer(const std::vector<std::string_view>& Arguments) {
     beammp_info("BeamMP Server v" + Application::ServerVersionString());
     Application::SetSubsystemStatus("Server", Application::Status::Starting);
+    std::mt19937_64 TimeSyncRandomDevice(std::random_device{}());
+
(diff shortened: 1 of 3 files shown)


src/TServer.cpp
```diff
@@ -123,6 +123,12 @@ TEST_CASE("GetPidVid") {
 TServer::TServer(const std::vector<std::string_view>& Arguments) {
     beammp_info("BeamMP Server v" + Application::ServerVersionString());
     Application::SetSubsystemStatus("Server", Application::Status::Starting);
+    std::mt19937_64 TimeSyncRandomDevice(std::random_device{}());
+
+    TimeSyncStart = std::chrono::steady_clock::now()
```
SaltySnail, line 128: This subtracts a random time between 0 - 24h from the current time right? I don't get the 2592000 comment, so maybe explain the thoughts behind this in a comment

SaltySnail (changes requested):
The rest looks good to me

src/TServer.cpp
WiserTixx: The comment was not intended to be committed. It will be gone in the next commit

SaltySnail (approved):
LGTM

